PR #1, opened 2020-08-24
Feature/convert custom data

blairlyons:
This PR changes the package name to simulariumio (shorter and less issues with underscores), sets the initial structure, adds functionality to convert spatiotemporal trajectory data from any source to .simularium JSON files, and adds functionality to add scatterplots or histograms to the data. I've included a jupyter notebook tutorial for how to use this (simulariumio/bin/Tutorial_custom.ipynb) which shows how to shape and convert random example data, both for spatial data and adding plots.

Notes on general structure:

- A main Converter object uses Reader objects of various types to read trajectory data for 3D display and metrics data for plots. (is 'Reader' the right name for this?)
- I've tried to set up the Readers to be extensible, so that different input sources and plot types can be added.
- Converter loads trajectory data in its constructor, since this data is required for all .simularium files. 
- For plot data, if it's already in Simularium format, it can be passed into the Converter constructor along with the trajectory data to be appended to the file. Otherwise, it can be added using the Converter.add_plot() function
- Right now I'm using objects with specific keys as arguments to the conversion functions, this might not be the best way to do this in Python? seems like getting the keys right could be error-prone...
- I'm also using strings to specify the engine or plot type, is there a way to make this less error-prone? something like an enum would be nice...
- This PR only deals with custom data, but soon I plan to add Readers for specific simulation engines (ReaDDy, Cytosim, PhysiCell). These will likely only require a path to the input data, since the data shape and file structure will be known. Is the current structure a good way to handle this? (i.e. a new Reader for each simulation engine, which takes an object as an argument, which likely only has one key for a file path), or should there be a way to directly input a file path for the non-custom Readers?

I'm hoping to get this merged soon so I can start adding the specific simulation engine support. 

This is my first Python package so I'm expecting you guys will have plenty of feedback :)


simulariumio/converter.py
```diff
@@ -0,0 +1,180 @@
...
+                        A numpy ndarray containing the radius 
+                        for each agent at each timestep
+                    subpoints: np.ndarray 
+                    (shape = [timesteps, agents, subpoints, 3]) (optional) 
+                        A numpy ndarray containing a list of subpoint position data 
+                        for each agent at each timestep. These values are 
+                        currently only used for fiber agents.
+                    plots : Dict[str, Any] (optional) 
+                        An object containing plot data already 
+                        in Simularium format
```
JacksonMaxfield, line 80: It may be better to make an entire documentation page about this format rather than including this in the docstring as it would allow more rich formatting. And then in the docstring simply say "go look at docs for format"

JacksonMaxfield: Would love @toloudis opinion though.

simulariumio/converter.py
```diff
@@ -0,0 +1,180 @@
...
+                'readdy' : outputs are from ReaDDy 
+                    (https://readdy.github.io/)
+                'physicell' : outputs are from PhysiCell 
+                    (http://physicell.org/)
+        """
+        traj_reader_class = self._determine_trajectory_reader(source_engine)
+        self._data = traj_reader_class().read(data)
+
+    @staticmethod
+    def _determine_trajectory_reader(source_engine: str = "custom") -> Type[Reader]:
```
JacksonMaxfield, line 98: I don't think you need the `Type` here... I don't _think_ :shrug:

simulariumio/converter.py
```diff
@@ -0,0 +1,180 @@
...
+    def write_JSON(self, output_path: str):
+        """
+        Save the data in .simularium JSON format at the output path
+
+        Parameters
+        ----------
+        output_path: str
+            where to save the file
+        """
+        with open("{}.simularium".format(output_path), "w+") as outfile:
```
JacksonMaxfield, line 179: I prefer fstrings so nitpick: `with open(f"{output_path}.simularium", "w+") as outfile:`

simulariumio/readers/custom_trajectory_reader.py
```diff
@@ -0,0 +1,174 @@
...
+        if "n_agents" not in data:
+            raise MissingDataError("n_agents")
+        if "viz_types" not in data:
+            raise MissingDataError("viz_types")
+        if "positions" not in data:
+            raise MissingDataError("positions")
+        if "types" not in data:
+            raise MissingDataError("types")
+        if "radii" not in data:
+            raise MissingDataError("radii")
```
JacksonMaxfield, line 115: It may be nicer to do something like:

```python
for field in ["box_size", "times", ...]:
    if field not in data:
        raise MissingDataError("viz_types")
```

Just another option if you prefer it that way.

simulariumio/readers/histogram_plot_reader.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        Return an object containing the data shaped for Simularium format
+        """
+
+        if "title" not in data:
+            raise MissingDataError("title")
+        if "xaxis_title" not in data:
+            raise MissingDataError("xaxis_title")
+        if "traces" not in data:
+            raise MissingDataError("traces")
```
JacksonMaxfield, line 28: You should look into Vega / vega-lite / altair for visualization specification: https://vega.github.io/vega-lite/docs/spec.html

Works in python and javascript so easy to convert between the two and may be good for this exact situation. Just a thought though. Don't know enough about your plot specification.

simulariumio/tests/readers/test_custom_trajectory_reader.py
```diff
@@ -0,0 +1,544 @@
...
+
+
+@pytest.mark.parametrize(
+    "trajectory, expected_data",
+    [
+        # 3 default agents (radius 5-10) at given positions for 5 frames,
+        # test string for plots
+        (
+            three_default_agents(),
+            {
```
JacksonMaxfield, line 18: FIRST: THANK YOU FOR TESTING

Second:
it may be good to move these large expected data objects into json files and simply read them in before asserting tests. I know others get frustrated when I have huge long parametrized tests because python prints out the entire parametrized set when reporting an error and so the test log becomes a mess.

simulariumio/converter.py
blairlyons: I copied how you guys did this in aicsimageio, but yes seems to be fine without it

JacksonMaxfield: brb. searching for the random `Type[...]` call in `aicsimageio` lmao

simulariumio/readers/custom_trajectory_reader.py
blairlyons: I decided to take this out, if a key is missing it will return a KeyError with the same info anyway

simulariumio/tests/readers/test_custom_trajectory_reader.py
blairlyons: I tried doing this but had issues with parsing, even though it was the same object in the file. Maybe you can help me figure that out, otherwise I'm just going to leave it.

simulariumio/converter.py
toloudis: I guess that changing the "Any" to some kind of python data structure definition would be my preferred thing.  But I don't know how expressive python's types can get, like can you specify "optional"?  And yes, this spec is the crux of the whole thing so it really needs to be easily findable in documentation.  But in my opinion, I think it would be easier to keep up-to-date with changes if it stays here near where it is used in the code.  

Also, it is not clear what the `str` key represents for the passed-in data dict.  Also, what does "loaded data or path to data" mean?

setup.py
```diff
@@ -63,23 +65,23 @@
     description="Simularium Conversion helps convert ssimulation outputs to the format consumed by the Simularium viewer.",
     entry_points={
         "console_scripts": [
-            "my_example=simularium_conversion.bin.my_example:main"
+            "my_example=simulariumio.bin.my_example:main"
```
toloudis, line 68: Do you want to name one of the bin scripts "my_example" ?

toloudis:
For file / directory structure this looks ok. How are users going to use this code?  Is the my_example supposed to be the canonical documentation that users will start with?

toloudis:
I see the python notebook now.  Makes sense to me.

toloudis:
There might be a way to write the Converter api a little more cleanly/explicitly.  Are most users going to be handing it a trajectory file or passing in the raw data by themselves?

simulariumio/converter.py
blairlyons: I think keeping the documentation near the code is a great idea FWIW.

The `str` is the key, how should I specify that? I'm trying to list them here...

by "loaded data or path to data" I mean that this object passed in might have numpy arrays of data in it, or it might have a filepath to where to load the data. Basically, for custom data like this, a user will have to shape their data into numpy arrays, but for specific simulation engines we support (which I'll be adding shortly) we'll know their output format so can load and shape the data for the user. Should I say this differently? Or just not say it since the keys indicate it?

setup.py
blairlyons: I did the example in jupyter notebook, should I add my_example.py too?

blairlyons:
It's hard to say what most users will do, I think passing in raw data could be just as common as passing in a specific engine's trajectory files. As we know, many people write their own simulators rather than using existing.

PR #2, opened 2020-09-12
Feature/convert cytosim data

blairlyons:
This PR adds the functionality to read CytoSim output files and to optionally visualize fiber points as spheres. This also updates to the v1 .simularium JSON format, and I split the plots tutorial out of the custom tutorial jupyter notebook so they're easier to find.


simulariumio/readers/cytosim_trajectory_reader.py
```diff
@@ -0,0 +1,377 @@
...
+
+###############################################################################
+
+
+class CytosimTrajectoryReader(TrajectoryReader):
+    def _ignore_line(self, line: str) -> bool:
+        """
+        if the line doesn't have any data, it can be ignored
+        """
+        return len(line) < 1 or line[0:7] == "warning" or "report" in line
```
EricJIsaac, line 25: Are these all the options for non-data lines in the Cytosim files? (e.g. there is no info/error to accompany the warning?)
Also, I'm guessing there isn't a way to directly identify data lines as opposed to disqualifying other line types?

EricJIsaac (approved):

simulariumio/readers/cytosim_trajectory_reader.py
blairlyons: yeah this is based on what I've seen in the test files I've been working with, there's no documented standard that I've found

PR #4, opened 2020-09-18
Feature/convert readdy data

blairlyons:
This PR adds the ability to read ReaDDy output files, with options to filter out or group specific particle types. Included is a ReaDDy example jupyter notebook and tests.


EricJIsaac (approved):

simulariumio/readers/readdy_trajectory_reader.py
```diff
@@ -0,0 +1,199 @@
...
+            i += 1
+        # assign group ID to each particle of a type in the group
+        for t in range(len(agent_data["n_agents"])):
+            for n in range(int(agent_data["n_agents"][t])):
+                readdy_id = agent_data["type_ids"][t][n]
+                if readdy_id in group_mapping:
+                    agent_data["type_ids"][t][n] = group_mapping[readdy_id]
+        return (agent_data, type_mapping)
+
+    def read(self, data: Dict[str, Any]) -> Dict[str, Any]:
```
toloudis, line 141: It would go a long way to documenting this api if you could change these `Any`s (both the input and the return value) to some kind of more specific type definition.  I don't know how feasible that is, or if this is part of the api that users would need to worry about.

toloudis (approved):

simulariumio/readers/readdy_trajectory_reader.py
blairlyons: there's going to be different types of data in this `data` object for different simulators, often fields won't be reused, for others they might be in a different format (see https://github.com/allen-cell-animated/simulariumio/blob/8f07c2c1fcd5a49c3f22e6f6b5dacb48be5359a6/simulariumio/converter.py#L49 for all the different simulators). I could try defining types for them, but I'm not sure how to handle inheritance with those type definitions. It seems like there isn't much overlap, so the definitions would be pretty complex and I'm not sure how useful they would be. Is there another way I should do this?

toloudis: When I see that docstring in converter.py showing all the different data for different simulators, it makes me think you need something like separate ReaddyConverter, PhysiCellConverter, CytosimConverter classes.  The generic Converter could discern the type it is dealing with and then forward the work to those classes.

PR #5, opened 2020-09-21
Feature/convert physicell data

blairlyons:
This PR adds the ability to read PhysiCell output files and assigns cells to agent types based on their cell type and current phase. Included is a PhysiCell example jupyter notebook and tests. 

PhysiCell reading depends on the pyMCDS.py script from the PhysiCell team, which is unfortunately just an example script and not part of a package. I added a copy of it in simulariumio/dep/ and added a comment with a link to the file on github. If there's a better way to handle this I'd love to improve it.


simulariumio/dep/pyMCDS.py
```diff
@@ -0,0 +1,548 @@
+import xml.etree.ElementTree as ET
+import numpy as np
+import pandas as pd
+import scipy.io as sio
+import sys
+import warnings
+from pathlib import Path
+
+"""
+Source: https://github.com/PhysiCell-Tools/python-loader/blob/master/pyMCDS.py
```
toloudis, line 10: To properly attribute this you might want to get the actual commit version url instead of the "master" one, since it might change over time.

EricJIsaac:
Are the files under  `simulariumio/tests/data/physicell/output/` input files for the tests? Or results generated by the tests?
If the second, they should probably be git-ignored

EricJIsaac (approved):

toloudis (approved):

PR #6, opened 2020-09-23
Feature/file format doc

blairlyons:
this PR adds documentation of the .simularium JSON format. I put it in a markdown file at the root of the repo, open to suggestions if there's a better place to put it.

### DIFF
--- a/file_format.md
+++ b/file_format.md
@@ -80,7 +80,7 @@ JSON files accepted by the simularium-viewer contain the following data in JSON 
                 "mesh" : "agent1.obj"  // optional
             },
             "2" : {
-                "name" : "agent2",     // optional
+                "name" : "agent2",
                 "mesh" : "agent2.obj"  // optional
             },
             ...
--- a/file_format.md
+++ b/file_format.md
@@ -1,7 +1,7 @@
 # .simularium JSON Format
 
 JSON files accepted by the simularium-viewer contain the following data in JSON format:
-* trajectory info
+* **trajectory info**
   * timeStepSize - the amount of time that passes in the simulation for each time step
   * totalSteps - the total number of time steps, or frames, in the simulation trajectory
   * size X, Y, Z - size of the bounding volume for the simulation. for now only rectangular prisms are supported, and the width in X, Y, and Z dimensions is provided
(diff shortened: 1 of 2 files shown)


toloudis:
We have it documented in typescript also, in the viewer.  That is a definitive "if this is broken nothing will work" version of the file format.  Because of that, I would compare what you have here with that one.

I wonder if we can auto-gen some of this from the viewer's type definitions, or produce a more formal type definition here.. like one of these https://json-schema.org/

toloudis (approved):

file_format.md
```diff
@@ -0,0 +1,193 @@
+# .simularium JSON Format
+
+JSON files accepted by the simularium-viewer contain the following data in JSON format:
+* **trajectory info**
+  * timeStepSize - the amount of time that passes in the simulation for each time step
```
toloudis, line 5: version is missing.  This doc is documenting version 1 and should make that clear.

blairlyons: whoops

blairlyons:
> We have it documented in typescript also, in the viewer. That is a definitive "if this is broken nothing will work" version of the file format. Because of that, I would compare what you have here with that one.

@toloudis VisDataMessage and CachedObservables don't have "version" in the viewer, is that ok?
https://github.com/allen-cell-animated/simularium-viewer/blob/f5b870b0d597a23a9708dcb5aa7ad72bbce19cfd/src/simularium/types.ts#L7

toloudis:
also on the subject of generating docs for data types:  http://typedoc.org/

toloudis:
> > We have it documented in typescript also, in the viewer. That is a definitive "if this is broken nothing will work" version of the file format. Because of that, I would compare what you have here with that one.
> 
> @toloudis VisDataMessage and CachedObservables don't have "version" in the viewer, is that ok?
> https://github.com/allen-cell-animated/simularium-viewer/blob/f5b870b0d597a23a9708dcb5aa7ad72bbce19cfd/src/simularium/types.ts#L7

I'm not sure if that is ok!   The interfaces in the viewer code that use it might not yet be using the version field. We should put it in there, tho, if it's really there.

file_format.md
```diff
@@ -0,0 +1,197 @@
...
+                "pdb" : "agent1.pdb",  // optional
+                "mesh" : "agent1.obj"  // optional
+            },
+            "1" : {
+                "name" : "agent1#bound",
+                "pdb" : "agent1.pdb",  // optional
+                "mesh" : "agent1.obj"  // optional
+            },
+            "2" : {
+                "name" : "agent2",     // optional
```
EricJIsaac, line 83: Is this optional?

EricJIsaac (changes requested):

file_format.md
blairlyons: it does fallback to ID, but I didn't mean to add optional to this one

EricJIsaac (approved):

PR #7, opened 2020-09-25
Feature/test uniqueIDs

blairlyons:
This PR adds a check that no agents have the same unique ID at any timestep, and it also fixes bugs in Cytosim reading that were causing IDs to overlap.

### DIFF
--- a/simulariumio/converter.py
+++ b/simulariumio/converter.py
@@ -256,6 +256,8 @@ class Converter:
         """
         traj_reader_class = self._determine_trajectory_reader(source_engine)
         self._data = traj_reader_class().read(data)
+        if not self._agent_ids_are_unique_per_frame():
+            raise Exception("agent unique IDs are not unique!")
 
     @staticmethod
     def _determine_trajectory_reader(source_engine: str = "custom") -> [Reader]:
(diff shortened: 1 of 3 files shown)


simulariumio/converter.py
```diff
@@ -279,6 +280,35 @@ def _determine_plot_reader(plot_type: str = "scatter") -> [Reader]:
...
+                        get_n_subpoints = False
+                        continue
+                    # there should be a unique ID at this index, check for duplicate
+                    uid = data[i]
+                    if uid in uids:
+                        raise Exception(
+                            f"found duplicate ID {uid} in frame {t} at index {i}"
+                        )
+                    uids.append(uid)
+                    next_uid_index += 9
```
toloudis, line 308: Where does this 9 come from? Is this something like the number of numeric entries per agent?  If we ever change the data format, this will have to change. It should at least be a symbolic global constant like `OFFSET_TO_NEXT_UID_V1=9 `but even better it could be the length of a list of all the per-agent data values `len(v1_data_struct=[X,Y,Z,RX,RY,RZ, ...])`

toloudis: Also, less important for now, but since each frame is independent, maybe this function could act only on a single frame t at a time, and could run inside the read() function or in the write() function?   I suppose this would be more of an optimization for very large trajectories, to not have to loop over all the data an extra time.

simulariumio/converter.py
```diff
@@ -279,6 +280,35 @@ def _determine_plot_reader(plot_type: str = "scatter") -> [Reader]:
...
+        For each frame, check that none of the unique agent IDs overlap
+        """
+        bundleData = self._data["spatialData"]["bundleData"]
+        for t in range(len(bundleData)):
+            data = bundleData[t]["data"]
+            next_uid_index = 1
+            uids = []
+            get_n_subpoints = False
+            for i in range(len(data)):
+                if i == next_uid_index:
```
EricJIsaac, line 294: This block could be accomplished in one loop - not strictly needed but easier to read at a glance imo:

e.g. while(i < len(data)) {
... assign directly to i instead of next_uid_index;
}

simulariumio/converter.py
EricJIsaac, line 308: If the # of sub-points were directly read, and the iterator assigned to the next uid (similar to line 298 but w/o the extra loop iteration), the structure of the code would better reflect the flow, making it easier to understand at a glance imo.

EricJIsaac (approved):

PR #13, opened 2020-10-19
Fix/move optional dependencies to their own directories

blairlyons:
moved specific engine related files to their own directories so that they don't cause issues if all the dependencies are not installed. This is an overall improvement, but also a fix for a windows user who cannot install ReaDDy

- [ ] Name your pull request _your-development-type/short-description_. Ex: _feature/read-tiff-files_
- [ ] Provide relevant tests for your feature or bug fix.

Thanks for contributing!


simulariumio/physicell/physicell_converter.py
```diff
@@ -6,12 +6,13 @@
 from pathlib import Path
 
 import numpy as np
-from .dep.pyMCDS import pyMCDS
+from ..dep.pyMCDS import pyMCDS
```
toloudis, line 9: is pyMCDS a physicell-specific dependency?

toloudis (approved):

simulariumio/physicell/physicell_converter.py
blairlyons: yes it is. should I move it to that directory?

PR #16, opened 2020-12-10
fix/floating point precision for timestep

blairlyons:
This is a fix for floating point precision errors in the timestep that were a problem for the front end time bar's calculations (Dan is adding a check there too). 

Thanks for contributing!

### DIFF
--- a/simulariumio/converter.py
+++ b/simulariumio/converter.py
@@ -223,6 +223,10 @@ class Converter:
         return True
 
     @staticmethod
+    def _format_timestep(number: float):
+        return float("%.4g" % number)
+
+    @staticmethod
     def _determine_plot_reader(plot_type: str = "scatter") -> [PlotReader]:
         """
         Return the plot reader to match the requested plot type
--- a/simulariumio/custom_converter.py
+++ b/simulariumio/custom_converter.py
@@ -41,7 +41,7 @@ class CustomConverter(Converter):
         totalSteps = len(input_data.agent_data.times)
         traj_info = {
             "version": 1,
-            "timeStepSize": (
+            "timeStepSize": Converter._format_timestep(
                 float(input_data.agent_data.times[1] - input_data.agent_data.times[0])
                 if totalSteps > 1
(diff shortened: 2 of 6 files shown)


schoinh (commented):
I can see the custom converter test is trying to compare converted data with expected data, but where can I see the expected data?

simulariumio/converter.py
```diff
@@ -222,6 +222,10 @@ def _check_agent_ids_are_unique_per_frame(self) -> bool:
                 get_n_subpoints = True
         return True
 
+    @staticmethod
+    def _format_timestep(number: float):
+        return float("%.4g" % number)
```
toloudis, line 227: Is 4 decimal places enough precision for the time step?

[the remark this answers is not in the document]
EricJIsaac: This will use string formatting to round a float to 4 significant figures in python
https://stackoverflow.com/a/3411731

blairlyons: I think so, we're only displaying 3 sig figs in the time bar currently, I've never seen a simulation timestep with more than 2 sig fig precision

blairlyons:
> I can see the custom converter test is trying to compare converted data with expected data, but where can I see the expected data?

it's at the top of test_custom_converter.py, there's 3 tests parameterized there, the expected data for each start at lines 18, 329, and 701

EricJIsaac (approved):

simulariumio/converter.py
toloudis: this is the trouble with passing through a text file format first.  we have to round things off before it even gets to the player. in a perfect world everything would be a literal float until it's time to display something and THEN it gets rounded.

toloudis (approved):

simulariumio/converter.py
EricJIsaac: Following the arraybuffer update, that will be an option. Whether or not we choose to implement this, it would be wise to support the text version despite this particular limitation.

PR #18, opened 2021-02-03
Feature/time units

blairlyons:
This PR adds a timeUnitFactorSeconds to allow users to control the time units. I'm using a float timeUnitFactorSeconds rather than a string timeUnit like we had discussed because it matches the current implementation for spatial units. 

We also need to discuss how to handle this new version on the front end, is it time to increment the version of trajectoryInfo or can we just check for the existence of this field for now and wait to increment after other imminent changes?

- [ ] Name your pull request _your-development-type/short-description_. Ex: _feature/read-tiff-files_
- [ ] Provide relevant tests for your feature or bug fix.

Thanks for contributing!


examples/Tutorial_custom.ipynb
```diff
@@ -47,18 +47,22 @@
...
-    "* **spatialUnitFactorMeters** : *float*\n",
-    "    * A float multiplier needed to convert spatial data (including positions, radii, and box size) to meters\n",
     "* **box_size** : *np.ndarray (shape = \\[3\\])*\n",
     "    * A numpy ndarray containing the XYZ dimensions of the simulation bounding volume\n",
     "* **agent_data** : `AgentData`\n",
     "    * An object containing data for each agent at each timestep\n",
+    "* **time_unit_factor_seconds** : *float (optional)*\n",
+    "    * A float multiplier needed to convert temporal data (e.g. timeStepSize) to seconds\n",
+    "    * Default: 1.0 (seconds)\n",
+    "* **spatialUnitFactorMeters** : *float (optional)*\n",
```
schoinh, line 57: Does this need to be `spatial_unit_factor_meters` instead? Seeing snake case elsewhere

examples/Tutorial_physicell.ipynb
```diff
@@ -79,7 +82,7 @@
     "\n",
     "example_data = PhysicellData(\n",
     "    box_size=np.array([box_size, box_size, 100.0]),\n",
-    "    timestep=360.0,  # seconds\n",
+    "    timestep=360000.0,\n",
```
schoinh, line 85: Is this supposed to represent 360 seconds (360000 ms)? Then shouldn't it be `timestep=360.0`. since we have the `time_unit_factor_seconds` specified as 1e-3?

simulariumio/data_objects/custom_data.py
```diff
@@ -34,19 +36,25 @@ def __init__(
...
             An object containing data for each agent
             at each timestep
+        time_unit_factor_seconds : float (optional)
+            A float multiplier needed to convert temporal data
+            (e.g. timeStepSize) to seconds
+            Default: 1.0
+        spatial_unit_factor_meters : float (optional)
+            A float multiplier needed to convert spatial data
+            (including positions, radii, and box size) to meters
+            Default: 1.0
```
schoinh, line 52: Would it be worth giving the example of "1e-9 = nanoseconds" here again for clarity?

simulariumio/tests/converters/test_physicell_converter.py
```diff
@@ -14,14 +14,16 @@
         (
             PhysicellData(
                 box_size=np.array([1000.0, 1000.0, 100.0]),
-                timestep=360.0,
+                timestep=360000.0,
```
schoinh, line 17: Same question in this file, should this be 360?

schoinh (approved):

toloudis:
Are these values such as 1e-9 ever going to be actually multiplied against time values?  Precision can become an issue.  I'm not sure I understand why the float representation is preferred against the string representation for units.

schoinh:
> Are these values such as 1e-9 ever going to be actually multiplied against time values? Precision can become an issue. I'm not sure I understand why the float representation is preferred against the string representation for units.

Yeah, if I'm understanding your question correctly they will be multiplied against the time values. I'm starting to wonder if it would be better for me to get a new test data file to develop with first, and when I'm sure that's the format I want, we update simulariumio? (ie convert this PR to draft for now?)

EricJIsaac (approved):

EricJIsaac:
> Are these values such as 1e-9 ever going to be actually multiplied against time values? Precision can become an issue. I'm not sure I understand why the float representation is preferred against the string representation for units.

Would this make a difference given the output here is JSON? (are precision errors being observed in the output files or further down the line?)

schoinh:
> > Are these values such as 1e-9 ever going to be actually multiplied against time values? Precision can become an issue. I'm not sure I understand why the float representation is preferred against the string representation for units.
> 
> Would string-vs-float make a difference given the output here is JSON? (are precision errors being observed in the output files or further down the line?)

@EricJIsaac by "string units" we mean `unit: "ns"` as opposed to `timeUnitFactorSeconds: 1e-9`, sorry that was confusing

blairlyons:
We already had this discussion about how to represent units. For the spatial units, we might multiply the factor by the distances and for at least one of my test trajectories last year (I think listeria?) I needed something other than an order of magnitude (i.e. 5e6). I think it would be really confusing if the time and spatial unit factors were handled differently. Can the front end have a function that gets a string unit from the factor? In python you could use pint, it just seems like fairly simple functionality?

schoinh:
> We already had this discussion about how to represent units. For the spatial units, we might multiply the factor by the distances and for at least one of my test trajectories last year (I think listeria?) I needed something other than an order of magnitude (i.e. 5e6). I think it would be really confusing if the time and spatial unit factors were handled differently. Can the front end have a function that gets a string unit from the factor? In python you could use pint, it just seems like fairly simple functionality?

Getting a string unit from the factor would be straightforward if the numbers were guaranteed to be nice, for example the time values range from 0 to 15 and the factor supplied is 1e9. Then converting this to 0 to 15 ns would be trivial. But if we had an input like this: time values range from 300 to 600 and the factor is 5e6, then I think I'd want to do some math which might include multiplying the time values by the factor. I think doing this type of math might have more consequence for time units compared to spatial units, because the calculation with the spatial unit is done once when the bounding box is drawn, but with time values we might have to convert the timestamp for every frame that comes in if the factor is not straightforward (e.g. 5e6)??? Maybe/hopefully there is a way to avoid this, I just can't think of one off the top of my head.

I just found a feature of the [math.js](https://mathjs.org/docs/datatypes/units.html#prefixes) library (which we don't have installed currently) that seems like it might do what Python's pint does, but again I'm not sure it would be super helpful for a weird case like the above with a factor of 5e6.

I can roll with this time unit factor and just try developing with it. But I'd like to hear @toloudis 's thoughts first since he's the one who first brought this up 😄

toloudis:
> > We already had this discussion about how to represent units. For the spatial units, we might multiply the factor by the distances and for at least one of my test trajectories last year (I think listeria?) I needed something other than an order of magnitude (i.e. 5e6). I think it would be really confusing if the time and spatial unit factors were handled differently. Can the front end have a function that gets a string unit from the factor? In python you could use pint, it just seems like fairly simple functionality?
> 
> Getting a string unit from the factor would be straightforward if the numbers were guaranteed to be nice, for example the time values range from 0 to 15 and the factor supplied is 1e9. Then converting this to 0 to 15 ns would be trivial. But if we had an input like this: time values range from 300 to 600 and the factor is 5e6, then I think I'd want to do some math which might include multiplying the time values by the factor. I think doing this type of math might have more consequence for time units compared to spatial units, because the calculation with the spatial unit is done once when the bounding box is drawn, but with time values we might have to convert the timestamp for every frame that comes in if the factor is not straightforward (e.g. 5e6)??? Maybe/hopefully there is a way to avoid this, I just can't think of one off the top of my head.
> 
> I just found a feature of the [math.js](https://mathjs.org/docs/datatypes/units.html#prefixes) library (which we don't have installed currently) that seems like it might do what Python's pint does, but again I'm not sure it would be super helpful for a weird case like the above with a factor of 5e6.
> 
> I can roll with this time unit factor and just try developing with it. But I'd like to hear @toloudis 's thoughts first since he's the one who first brought this up 😄

I didn't mean to reopen a huge can of worms.  My opinion is that it's safer and clearer to specify a unit from a well known enumerated list of units, and a number multiplier along with it.  So 5e3 becomes [5.0, "km"]. 

Maybe this is all too nitpicky and the single float value is fine.  Here's the weird case, when the exponent is "in-between":
If a user specifies 4e5, that's equivalent to 400km, right? Or is it 0.4 Mm? (We would have to convert the 5 into the nearest power that gives a recognizable string unit, and multiply the remaining factor against the 4?)

schoinh:
> > > We already had this discussion about how to represent units. For the spatial units, we might multiply the factor by the distances and for at least one of my test trajectories last year (I think listeria?) I needed something other than an order of magnitude (i.e. 5e6). I think it would be really confusing if the time and spatial unit factors were handled differently. Can the front end have a function that gets a string unit from the factor? In python you could use pint, it just seems like fairly simple functionality?
> > 
> > 
> > Getting a string unit from the factor would be straightforward if the numbers were guaranteed to be nice, for example the time values range from 0 to 15 and the factor supplied is 1e9. Then converting this to 0 to 15 ns would be trivial. But if we had an input like this: time values range from 300 to 600 and the factor is 5e6, then I think I'd want to do some math which might include multiplying the time values by the factor. I think doing this type of math might have more consequence for time units compared to spatial units, because the calculation with the spatial unit is done once when the bounding box is drawn, but with time values we might have to convert the timestamp for every frame that comes in if the factor is not straightforward (e.g. 5e6)??? Maybe/hopefully there is a way to avoid this, I just can't think of one off the top of my head.
> > I just found a feature of the [math.js](https://mathjs.org/docs/datatypes/units.html#prefixes) library (which we don't have installed currently) that seems like it might do what Python's pint does, but again I'm not sure it would be super helpful for a weird case like the above with a factor of 5e6.
> > I can roll with this time unit factor and just try developing with it. But I'd like to hear @toloudis 's thoughts first since he's the one who first brought this up 😄
> 
> I didn't mean to reopen a huge can of worms. My opinion is that it's safer and clearer to specify a unit from a well known enumerated list of units, and a number multiplier along with it. So 5e3 becomes [5.0, "km"].
> 
> Maybe this is all too nitpicky and the single float value is fine. Here's the weird case, when the exponent is "in-between":
> If a user specifies 4e5, that's equivalent to 400km, right? Or is it 0.4 Mm? (We would have to convert the 5 into the nearest power that gives a recognizable string unit, and multiply the remaining factor against the 4?)

I like the [5.0, "km"] idea. I think it prevents working with really small or really large numbers, gives the user flexibility (with the 5.0), gives the user/developer clarity ("km" is the unit meant to be displayed), and in the vast majority of cases the first number is going to be 1 which means we won't do any multiplying.

If we applied it to spatial units as well, then we won't have the issue of the user having to follow 2 different protocols for space units and time units.

simulariumio/data_objects/unit_data.py
```diff
@@ -0,0 +1,55 @@
...
+        magnitude: float (optional)
+            multiplier for values (in case they are not given in whole units)
+            Default: 1.0
+        """
+        # standardize and simplify units with pint
+        ureg = UnitRegistry()
+        self._quantity = magnitude * ureg(name)
+        self._quantity = self._quantity.to_compact()
+        self.magnitude = self._clamp_precision(self._quantity.magnitude)
+        self.name = str(self._quantity.units)
```
schoinh, line 40: Is there a reason to return the unabbreviated unit name here ("seconds" as opposed to "s")? We only display abbreviated units in the frontend currently, so it would be nice to be able to use the abbreviated string unit straight from the data.

simulariumio/data_objects/unit_data.py
```diff
@@ -0,0 +1,55 @@
...
+        name: str
+            unit name for values (we support this list
+            https://github.com/hgrecco/pint/blob/master/pint/default_en.txt)
+        magnitude: float (optional)
+            multiplier for values (in case they are not given in whole units)
+            Default: 1.0
+        """
+        # standardize and simplify units with pint
+        ureg = UnitRegistry()
+        self._quantity = magnitude * ureg(name)
```
schoinh, line 37: What happens here if a non-valid unit is passed in? Like `"kitties"`

simulariumio/data_objects/unit_data.py
```diff
@@ -0,0 +1,55 @@
...
+    ):
+        """
+        This object contains data for units
+
+        Parameters
+        ----------
+        name: str
+            unit name for values (we support this list
+            https://github.com/hgrecco/pint/blob/master/pint/default_en.txt)
+        magnitude: float (optional)
```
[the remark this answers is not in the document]
blairlyons: I got this from pint, the Python package for units. It has "quantities" that have "magnitude "and "units". Of course I'm using "name" instead of "units" because it seemed less redundant and should help indicate that it's a string. I don't feel that strongly about it though.

file_format.md
```diff
@@ -58,12 +63,14 @@ JSON files accepted by the simularium-viewer contain the following data in JSON
...
+        "version" : 2,
+        // time units
+        "timeUnitFactorSeconds" : 1e-6, // microseconds
         // time
         "timeStepSize" : 0.5,
         "totalSteps" : 1000,
-        // units
-        "spatialUnitFactorMeters" : 1e-9,
+        // spatial units
+        "spatialUnitFactorMeters" : 1e-9, // nanometers
```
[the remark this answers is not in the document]
blairlyons: good catch!

simulariumio/data_objects/unit_data.py
blairlyons: pint outputs unit strings unabbreviated, I can look into getting it to output the abbreviated units, but won't you be using a units library anyway?

simulariumio/data_objects/unit_data.py
blairlyons: it raises a pint.errors.UndefinedUnitError. I tried to add a test for that but couldn't get it to work

schoinh: I think it would be good to display an error message with a link to the list of acceptable units, but if others don't think it's necessary either I don't feel strongly about it :)

simulariumio/data_objects/unit_data.py
schoinh: I thought the utility of using a units library would be in screening out nonstandard units, which is being done here with pint. @toloudis and I [talked about this briefly on Slack](https://allencellscience.slack.com/archives/CFT511CE4/p1613161719045500), let me know if I'm missing something.

Would this work? `self.name = f'{self._quantity.units:~}'`
source: https://pint.readthedocs.io/en/0.10.1/tutorial.html#string-formatting

simulariumio/data_objects/unit_data.py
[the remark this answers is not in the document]
schoinh: Ah I see. Up to you.

simulariumio/tests/converters/test_readdy_converter.py
```diff
@@ -13,20 +14,28 @@
...
+                time_units=UnitData("ms", 1e-6),
+                spatial_units=UnitData("nm"),
             ),
             {
                 "trajectoryInfo": {
-                    "version": 1,
+                    "version": 2,
+                    "timeUnits": {
+                        "magnitude": 1.0,
+                        "name": "nanosecond",
```
EricJIsaac, line 31: Is this supposed to match the above time_unit ("ms")?

EricJIsaac (approved):

schoinh (commented):
Un-approving approval for ~last week's version

simulariumio/tests/converters/test_readdy_converter.py
blairlyons: yes but 1e-6 ms is 1 ns, I'm testing to make sure the units get simplified correctly

simulariumio/data_objects/unit_data.py
blairlyons: well you can't count on all data coming through this package, some people will write their own scripts to convert their data. So I think you still need to check that on the FE.

I changed everything to abbreviations though, that was my preference too just was being lazy :) however, in doing that I discovered that pint's abbreviation for micrometers is wrong 😱 it's "µ" rather than "µm". I wrote a workaround but now I'm wondering if we should put in a bug to pint for that. Seems like someone else would have noticed by now??

schoinh: Ah, ok, I didn't consider people converting data on their own. Thanks for abbreviating the units anyway :) 

What? I just Googled and found that µ is the abbreviation for "micron" before micron got replaced with micrometers..... you probably knew that, but anyway sounds like they need to update their stuff!

schoinh (approved):
Yay, thank you!!

simulariumio/data_objects/unit_data.py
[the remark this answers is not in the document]
toloudis: I would have gone with `timeUnit : str` and `timeScale : float` instead of a nested struct 😁

simulariumio/data_objects/unit_data.py
toloudis: To me, the unit representation is an identifier that can be used to look up the name of the unit, the order of magnitude of the base SI unit (i.e. 10^6 etc) and other info.  If that identifier happens to be a human readable string abbreviation, that can be output in the front end, that's great!  As long as it uniquely identifies the unit!

For the micro symbol, I wonder if we need to utf-8 encode it for json output or does  pint handle that.

PR #19, opened 2021-02-15
Feature/read and filter

blairlyons:
This PR adds functionality to read simularium JSON files back in and filter them to reduce the number of agents:
- The old base class, Converter, is now combined with CustomConverter (which is now the converter base class), since functions in CustomConverter are needed to read data back in, and then out after filtering
- A new FileConverter is added, which loads data from JSON, and AgentData has added functions to shape the loaded data back into the format needed for filtering
- A Filter base class is added, as well as a first filter for number of agents

Thanks for contributing!


simulariumio/custom_converter.py
```diff
@@ -1,22 +1,56 @@
...
 
 ###############################################################################
 
+SUPPORTED_PLOT_READERS = {
+    "scatter": ScatterPlotReader,
+    "histogram": HistogramPlotReader,
+}
+
+FILTERS = {
+    "every_nth_agent": EveryNthAgentFilter,
```
toloudis, line 45: can users make their own filters that don't live in this repo?

simulariumio/custom_converter.py
```diff
@@ -92,3 +119,276 @@ def _read(self, input_data: CustomData) -> Dict[str, Any]:
...
+                n_subpoints = int(agent_data.n_subpoints[t][n])
+                if n_subpoints > 0:
+                    # add subpoints to fiber agent
+                    subpoints = [3 * n_subpoints]
+                    for p in range(n_subpoints):
+                        for d in range(3):
+                            subpoints.append(agent_data.subpoints[t][n][p][d])
+                    local_buf[
+                        i
+                        + V1_SPATIAL_BUFFER_STRUCT.index("NSP") : i
```
toloudis, line 179: if V1_SPATIAL_BUFFER_STRUCT never changes, all these `.index` calls could be symbolic constants.  Or at least the hardcoded strings could be.

simulariumio/custom_converter.py
```diff
@@ -92,3 +119,276 @@ def _read(self, input_data: CustomData) -> Dict[str, Any]:
...
+        Parameters
+        ----------
+        params: List[FilterParams]
+            a list of filter parameters, one for each filter to be applied
+        """
+        box_size = self._data["trajectoryInfo"]["size"]
+        plot_data = self._data["plotData"]
+        agent_data = AgentData.from_simularium_data(self._data)
+        for i in range(len(params)):
+            filter_class = self._determine_filter(params[i].name)
```
toloudis, line 367: Perhaps passing in the Filter itself rather than its string name will make this more extensible

toloudis:
This is going to turn into a very useful toolset for manipulating trajectory data.  
I have a couple high level future suggestions.

1. have `applyfilter` return the new data out instead of modifying it in-place.  This will be less destructive but at a cost of memory.  (to save memory it might be possible to do the filtering at write-time but i don't know)

With the above change you could chain filters like this:  
mydataTransformed3Times = myorigdata.applyFilter(F1).applyFilter(F2).applyfilter(F3)
 in addition to passing them as a list 
applyFilter([F1, F2, F3])

2. Let users make their own filters without having to modify this repo.  Allow applyFilter() to take a Filter object or a lambda function or something.

3. Random thought, it might be interesting to have two flavors of Filters, one that runs per-frame (the filter function only takes one frame's worth of data and outputs one frame of data) and one that runs over all the frames.
So you would have a flow like this:
for each frame in trajectory:
  test against per-time filter to see if the frame is dropped or not
  if frame not dropped, apply per-frame filter and add the result to a new trajectory
 return new trajectory

toloudis (approved):

simulariumio/custom_converter.py
blairlyons: no they can't, but I should fix that

simulariumio/custom_converter.py
blairlyons: what would V1_SPATIAL_BUFFER_STRUCT look like then?

simulariumio/custom_converter.py
blairlyons: how should I pass both the filter and the parameters? Make the filter function itself one of the parameters? Or pass them both in?

blairlyons:
> This is going to turn into a very useful toolset for manipulating trajectory data.
> I have a couple high level future suggestions.
> 
>     1. have `applyfilter` return the new data out instead of modifying it in-place.  This will be less destructive but at a cost of memory.  (to save memory it might be possible to do the filtering at write-time but i don't know)
> 
> 
> With the above change you could chain filters like this:
> mydataTransformed3Times = myorigdata.applyFilter(F1).applyFilter(F2).applyfilter(F3)
> in addition to passing them as a list
> applyFilter([F1, F2, F3])
> 
>     1. Let users make their own filters without having to modify this repo.  Allow applyFilter() to take a Filter object or a lambda function or something.
> 
>     2. Random thought, it might be interesting to have two flavors of Filters, one that runs per-frame (the filter function only takes one frame's worth of data and outputs one frame of data) and one that runs over all the frames.
>        So you would have a flow like this:
> 
> 
> ```
> for each frame in trajectory:
>     test against per-time filter to see if the frame is dropped or not
>     if frame not dropped, apply per-frame filter and add the result to a new trajectory
>  return new trajectory
> ```
> 
> That last one is a half baked idea and I'm not sure it has merit as it is.

For 1, I didn't set it up to chain the filters because I translate the numpy arrays back into the buffer format at the end of `apply_filters()` so it's a lot more performant to do them all before I do that. But I could store the data in the numpy arrays instead, and only go back to the buffers in `write_JSON()`. Now that I think of that I'm not sure why I didn't do that already.

2 sounds good, for 3, I want to go back and add per frame functionality in general. That would be good to add too.

simulariumio/custom_converter.py
```diff
@@ -92,3 +118,263 @@ def _read(self, input_data: CustomData) -> Dict[str, Any]:
...
         }
         return simularium_data
+
+    @staticmethod
+    def _get_spatial_bundle_data_subpoints(
+        agent_data: AgentData,
+    ) -> List[Dict[str, Any]]:
+        """
+        Return the spatialData's bundleData for a simulation
+        of agents with subpoints, packing buffer with jagged data is slower
```
toloudis, line 128: I have often wondered if it would be more efficient to keep the jagged data separate.  Put subpoints in a different buffer or have them entirely at the end of the per-agent data.   It's not only faster to write to, but also faster to parse if the data is a fixed known size per agent.  Just thinking out loud (this gives me an idea to have a simplified fiber mode that doesn't need to create the full 3d tubes).

simulariumio/custom_converter.py
```diff
@@ -92,3 +118,263 @@ def _read(self, input_data: CustomData) -> Dict[str, Any]:
...
+                            local_buf[i + V1_SPATIAL_BUFFER_STRUCT.R_INDEX] = 0.5
+                            i += V1_SPATIAL_BUFFER_STRUCT.VALUES_PER_AGENT - 1
+                else:
+                    i += V1_SPATIAL_BUFFER_STRUCT.VALUES_PER_AGENT - 1
+            frame_data["data"] = local_buf.tolist()
+            bundle_data.append(frame_data)
+        return bundle_data
+
+    @staticmethod
+    def _get_spatial_bundle_data_no_subpoints(
```
toloudis, line 224: when do you use this no_subpoints version?

blairlyons: for some engines there's never any fibers (ReaDDy for example), so the _read_custom_data() function will use this version for that since the subpoints array will be None. I kept this because it seems like it could be a lot more performant in the case of no fibers? But haven't profiled that or anything

simulariumio/custom_converter.py
blairlyons: Yeah I've wondered that too and I think we've discussed it a little in the past, but definitely could discuss again

toloudis (approved):
My only concern is the name CustomData is really SimulariumTrajectoryData, right?  It's not really custom to anything, but it's the true simularium data format.  Am I understanding that correctly?  And a CustomConverter is really just a TrajectoryConverter?

blairlyons:
> My only concern is the name CustomData is really SimulariumTrajectoryData, right? It's not really custom to anything, but it's the true simularium data format. Am I understanding that correctly? And a CustomConverter is really just a TrajectoryConverter?

That's true, except `CustomData` isn't exactly in simularium format, it's still numpy arrays rather than the buffers. The names are leftover from when there was a `Converter` base class and `CustomConverter` was a derived class for non-engine specific data. But yes, I think I could change `CustomConverter` --> `TrajectoryConverter` and `CustomData` --> `TrajectoryData`?

toloudis:
> 
> But yes, I think I could change `CustomConverter` --> `TrajectoryConverter` and `CustomData` --> `TrajectoryData`?

To me, that change makes a lot of sense and makes the code a bit more understandable

simulariumio/data_objects/agent_data.py
```diff
@@ -95,3 +99,165 @@ def __init__(
         self.n_subpoints = n_subpoints
         self.subpoints = subpoints
         self.draw_fiber_points = draw_fiber_points
+        self.type_ids = type_ids
+
+    @staticmethod
+    def _get_buffer_data_dimensions(buffer_data: Dict[str, Any]) -> Tuple[int]:
+        """"""
+        bundleData = buffer_data["spatialData"]["bundleData"]
```
toloudis, line 107: i just noticed this bundleData should really be bundle_data to conform to Python variable naming conventions. I believe either `black` or `flake8` should catch these things and let you know about them

blairlyons: I always run black and I think flake8 is run as part of the build, but they don't catch this

toloudis: I noticed that too, the build runs both.  I could have sworn that in my editor i see those things reported but maybe they are different config settings.

simulariumio/tests/converters/test_cytosim_converter.py
```diff
@@ -23,7 +23,7 @@
                     "fibers": CytosimObjectInfo(
                         filepath="simulariumio/tests/data/cytosim"
                         "/3_fibers_3_frames/fiber_points.txt",
-                        agents={0: CytosimAgentInfo(name="fiber")},
+                        agents={0: CytosimAgentInfo(name="fiber", radius=1e-3)},
```
schoinh, line 26: Why was this change necessary?

schoinh (approved):
All of my comments are minor / for my curiosity :)

simulariumio/custom_converter.py
```diff
@@ -85,11 +100,15 @@ def _read(self, input_data: CustomData) -> Dict[str, Any]:
             "bundleSize": totalSteps,
         }
         if input_data.agent_data.subpoints is not None:
-            spatialData["bundleData"] = self._get_spatial_bundle_data_subpoints(
+            spatialData[
+                "bundleData"
+            ] = TrajectoryConverter._get_spatial_bundle_data_subpoints(
```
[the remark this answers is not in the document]
blairlyons: I don't think it matters? both the instance and the class have access to the static function and it seems like I can use either in testing. A python expert might know...

simulariumio/file_converter.py
```diff
@@ -0,0 +1,64 @@
...
+                "magnitude": spatial_units.magnitude,
+                "name": spatial_units.name,
+            }
+            data["trajectoryInfo"].pop("spatialUnitFactorMeters")
+            time_units = UnitData("s", 1.0)
+            data["trajectoryInfo"]["timeUnits"] = {
+                "magnitude": time_units.magnitude,
+                "name": time_units.name,
+            }
+            data["trajectoryInfo"]["version"] = 2
```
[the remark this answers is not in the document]
blairlyons: for this one I made it explicit because this part of the function is converting v1 -> v2, I'll need a new block for converting v2 -> v3 someday when the current version is 3

simulariumio/cytosim/cytosim_converter.py
```diff
@@ -43,18 +43,21 @@ def _ignore_line(self, line: str) -> bool:
...
     ) -> Tuple[List[int], int]:
         """
-        Parse a Cytosim fiber_points.txt output file to get the number
+        Parse a Cytosim output file containing objects
+        (fibers, solids, singles, or couples) to get the number
         of subpoints per agent per timestep
         """
         result = []
         t = -1
         s = 0
```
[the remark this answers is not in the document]
blairlyons: these are indices so I did this for brevity, e.g. `result.n_subpoints[t][n_other_agents + n] = s + 1` would otherwise be `result.n_subpoints[time][n_other_agents + agent] = subpoint + 1` over and over. I guess I could change it but would rather not change it for everything right now

simulariumio/filters/every_nth_agent_filter.py
```diff
@@ -0,0 +1,110 @@
...
+###############################################################################
+
+log = logging.getLogger(__name__)
+
+###############################################################################
+
+
+class EveryNthAgentFilter(Filter):
+    n_per_type_id: Dict[int, int]
+    default_n: int
```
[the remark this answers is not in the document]
blairlyons: yes, so other files in the project know about them and their types

simulariumio/tests/converters/test_cytosim_converter.py
blairlyons: because radii are multiplied by scale factor. I thought they were before this PR but anyway this makes sense and is needed for the test to pass

PR #21, opened 2021-03-04
Feature/basic filters

blairlyons:
This PR adds more basic filters on top of the reading and filtering functionality in #19 : every Nth subpoint, every Nth timestep, add agents, multiply time, reorder agents, transform spatial axes, and translate.

Also added a add_number_of_agents_plot() function in TrajectoryConverter, fixed an error in cytosim test data (times were [0.0, 0.5, 0.5] instead of [0.0, 0.5, 1.0]), and moved a scatter plot test fixture so it can be reused.

Thanks for contributing!


simulariumio/filters/add_agents_filter.py
```diff
@@ -0,0 +1,138 @@
...
+###############################################################################
+
+
+class AddAgentsFilter(Filter):
+    new_agent_data: AgentData
+
+    def __init__(self, new_agent_data: AgentData):
+        """
+        This filter adds the given agents
+        to each frame of the simulation
```
toloudis, line 26: maybe TrajectoryData could use an append_agents(AgentData) function?

toloudis: or even just a single append_agent that could be called in a loop

simulariumio/filters/multiply_time_filter.py
```diff
@@ -0,0 +1,76 @@
...
+    multiplier: float
+    apply_to_plots: bool
+
+    def __init__(
+        self,
+        multiplier: float,
+        apply_to_plots: bool = True,
+    ):
+        """
+        This filter multiplies time values
```
EricJIsaac, line 29: nit: similar to above, would be helpful to specify 'scales the time-stamp values for each simulation frame by a provided multiplier'

EricJIsaac (approved):

simulariumio/trajectory_converter.py
```diff
@@ -335,6 +337,39 @@ def add_plot(
...
+                type_name = type_mapping[str(int(agent_data.type_ids[t][n]))]["name"]
+                if "#" in type_name:
+                    type_name = type_name.split("#")[0]
+                if type_name not in n_agents:
+                    n_agents[type_name] = np.zeros_like(agent_data.times)
+                n_agents[type_name][t] += 1
+        self.add_plot(
+            ScatterPlotData(
+                title="Number of agents over time",
+                xaxis_title="Time (s)",
```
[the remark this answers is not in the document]
blairlyons: good catch! fixed it

blairlyons:
@toloudis I just updated to add the append_agents() method. I also noticed I wasn't copying the TrajectoryData correctly so the previous data was getting modified during filtering. I fixed that and added a test for it too

simulariumio/data_objects/agent_data.py
```diff
@@ -263,3 +268,108 @@ def from_buffer_data(cls, buffer_data: Dict[str, Any]):
             draw_fiber_points=False,
             type_ids=type_ids,
         )
+
+    def append_agents(self, new_agents: AgentData):
+        """
+        Concatenate the new AgentData with the current data,
+        generate new unique IDs and type IDs as needed
```
toloudis, line 275: one question:  what if I want to append the same N agents to every time step (e.g. merging two whole trajectories that have same number of time steps) -- how can i ensure the agent ids from new_agents will stay the same across time if this is reassigning ids?   
Maybe this is a consideration for later but I bet it could come up easily.

toloudis: oh, i see that new_agents is a whole trajectory, not just one time step

toloudis: so new_agents is required to have same number of time steps as self?

toloudis (approved):

simulariumio/data_objects/agent_data.py
blairlyons: yeah it is. I want to add per frame functionality in a later PR, there's a lot of stuff that needs to be implemented and it makes sense to me to do it all at once?

blairlyons: if you wanted to just add to some frames, for now you could have no agents in those frames in the added trajectory

PR #22, opened 2021-03-05
Feature/default camera transform

blairlyons:
This PR adds default camera transform data, including camera position XYZ, look at position XYZ, up vector XYZ, and field of view angle in degrees. Since all the converters need the ability to set this data, along with box_size and scale_factor, I added a new MetaData object to store all of these per-trajectory global parameters, and a CameraData object to store all the camera-specific values.

Key files:
- file_format.md : added specification for camera defaults
- data_objects/meta_data.py, camera_data.py : new data structures
- everything else is updates and tests for the new data structures

Thanks for contributing!


EricJIsaac (approved):

simulariumio/data_objects/trajectory_data.py
```diff
@@ -38,9 +37,9 @@ def __init__(
 
         Parameters
         ----------
-        box_size : np.ndarray (shape = [3])
-            A numpy ndarray containing the XYZ dimensions
-            of the simulation bounding volume
+        meta_data : MetaData
```
schoinh, line 40: Doesn't it seem odd to have `time_units` and `spatial_units` separate from `meta_data`....... since they are technically metadata...?  Not that I _want_ to fiddle with changes to the unit data shape in simularium-viewer again... 🙈

schoinh: I wonder if it might be appropriate to change `meta_data`/`MetaData` to `scene_data`/`SceneData` or something like that? Metadata is very broad. But if I'm the only person who thinks this, don't worry about it haha

blairlyons: some packages don't need the user to specify units, for example Cytosim is always seconds and micrometers, and this metadata object is used by all converters. I think it would be confusing if a Cytosim user thought they could enter units but then they were overridden by the Cytosim converter. And it seems fine to have units as fields on TrajectoryData?

schoinh: Oh I see, okay.

schoinh (approved):

PR #24, opened 2021-03-31
Feature/convert springsalad

blairlyons:
This PR adds conversion from SpringSaLaD (https://vcell.org/ssalad). I'm using pandas to store the parsed data. I think it will be a good idea to use pandas to store data more in general, but that will be a future PR.

Key files:
- setup.py : added pandas requirement
- data_objects/agent_data.py : added from_dataframe() method to create AgentData from pandas DataFrame (in the future maybe AgentData should store its data in a DataFrame?)
- springsalad/springsalad_converter.py, springsalad_data.py : SpringSaLaD conversion (if you have any suggestions for the pandas implementation, please lmk, I'm a pandas n00b)

I also added the same display_names parameter from SpringsaladData to the Smoldyn conversion, since I've already run into issues with the way I originally did that, and this is more flexible. And fixed a few other small things and added None checks.

Thanks for contributing!


EricJIsaac:
The Jupyter notebooks seem reasonably generatable - this may be worth looking into depending on how many more packages we'd like to add

EricJIsaac (approved):

simulariumio/springsalad/springsalad_converter.py
```diff
@@ -0,0 +1,108 @@
...
+        for line in springsalad_data:
+            cols = line.split()
+            if "xsize" in line:
+                box_size[0] = input_data.scale_factor * 2 * float(cols[1])
+            if "ysize" in line:
+                box_size[1] = input_data.scale_factor * 2 * float(cols[1])
+            if "z_outside" in line:
+                box_size[2] += input_data.scale_factor * 2 * float(cols[1])
+            if "z_inside" in line:
+                box_size[2] += input_data.scale_factor * 2 * float(cols[1])
```
toloudis, line 65: i don't know what this code is doing but it looks fishy

toloudis: Also why is everything being doubled?

blairlyons: SpringSaLaD output files have bounding volume parameters like:

`xsize	50.0
ysize	50.0
z_outside	10.0
z_inside	90.0`

and by trial and error I found that they're the box "radius" rather than dimensions, that's why they are multiplied by 2. And then the Z radius is z_inside + z_outside. It seems weird to me, but it's how the engine does it?

toloudis: could they mean "inside the cell" and "outside the cell"?

toloudis (approved):

simulariumio/springsalad/springsalad_converter.py
blairlyons: yes I'm almost sure that's what they mean

PR #27, opened 2021-05-10
Feature/convert mcell binary

blairlyons:
Problem
=======
Simularium can't read simulation outputs from MCell.

Solution
========
This adds conversion for MCell binary visualization files. I used MCell's own python implementation of reading from the binary files as a reference. MCell also has an ascii file format, we can add support later but it seems like most outputs are currently saved to binary. Also, the MCell binary is currently missing unique IDs. They plan to add this to the next version of the format, and until then I'm assigning unique IDs based on agent index.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

Change summary:
---------------
* Read MCell binary files using python array.array
* Go through the data twice in order to create data structures with required length for optimization
* Calculate euler angle rotations randomly oriented around the MCell normals, and check that they match the Blender visualization
* Add tests and example jupyter notebook

Keyfiles:
-----------------------
1. simulariumio/mcell/mcell_converter.py - implemented MCell conversion
2. simulariumio/mcell/mcell_data.py - data a user needs to provide for MCell conversion
2. there's some small fixes and reorganization in simulariumio/data_objects/agent_data.py and simulariumio/filters/transform_spatial_axes_filter.py, artifacts of how I split adding rotations into a separate PR


schoinh (approved):

simulariumio/mcell/mcell_converter.py
```diff
@@ -0,0 +1,347 @@
...
+                    is_surface_mol = is_surface_mol[0] == 1
+                    n_data = array.array("I")
+                    n_data.fromfile(mol_file, 1)
+                    n_data = n_data[0]
+                    data = array.array("f")
+                    data.fromfile(mol_file, n_data)
+                    if is_surface_mol:
+                        data.fromfile(mol_file, n_data)
+                    total_mols += int(n_data / 3.0)
+                except EOFError:
```
toloudis, line 157: this EOFError check seems clunky - the file has no indication of how many times you'll be able to run this loop?

toloudis (approved):

simulariumio/mcell/mcell_converter.py
blairlyons: yeah, the binary data doesn't include that https://cnl.salk.edu/~ahusar/mcell4_documentation/file_formats.html#molecule-visualization-data and this is how they implemented reading it 🤷‍♀️

PR #28, opened 2021-05-14
Fix/overlapping type ids in add agent filter

blairlyons:
Problem
=======
AddAgentFilter outputs overlapping type IDs if any of the new agents have the same type ID as an existing agent.

Solution
========
Only combine the AgentDatas' types list, which is the list of string type names for each agent at each timestep. Instead of generating and then also combining the type_ids array, generate it after creating the new types list.

## Type of change

Change summary:
---------------
* In AgentData's append_agents() function, regenerate the type_ids array from the new types list


simulariumio/data_objects/agent_data.py
```diff
@@ -412,12 +403,12 @@ def append_agents(self, new_agents: AgentData):
...
                     used_uids.append(uid)
                 unique_ids[t][i] = new_uids[raw_uid]
-                type_ids[t][i] = new_agents.type_ids[t][n]
                 types[t].append(new_agents.types[t][n])
                 i += 1
         self.unique_ids = unique_ids
         self.types = types
-        self.type_ids = type_ids
+        self.type_ids, tm = AgentData.get_type_ids_and_mapping(self.types)
+        self.type_mapping = None
```
EricJIsaac, line 411: Not related to stated problem, just a curiosity -
Why is the type-mapping here ignored and not assigned? Would something not behave as expected if it were assigned?

EricJIsaac (approved):

toloudis (approved):

simulariumio/data_objects/agent_data.py
blairlyons: `AgentData.type_mapping` is leftover from previous ways of handling this. I'll just remove it because it's not used anymore. Now I always generate it right before writing to JSON. `AgentData.types`, a list of string type names, is the source of truth for types now, and type IDs and type mapping are generated from it.

PR #29, opened 2021-06-16
Feature/default camera fov 75

blairlyons:
Problem
=======
The Simularium Viewer's default camera FOV is 75 and SimulariumIO's was 50.
Resolves https://aicsjira.corp.alleninstitute.org/browse/AGENTVIZ-1412 

Solution
========
Update SimulariumIO's default FOV to 75.

## Type of change

Change summary:
---------------
* change default FOV in CameraData
* update tests and documentation

Keyfiles:
-----------------------
1. simulariumio/data_objects/camera_data.py - update default FOV for a CameraData, which holds camera settings
2. simulariumio/file_converter.py - update default FOV when older file versions are read in
3. file_format.md - update file format documentation
4. jupyter notebooks and tests - update FOV

### DIFF
--- a/examples/Tutorial_custom.ipynb
+++ b/examples/Tutorial_custom.ipynb
@@ -82,7 +82,7 @@
     "    * Default: np.array(\\[0.0, 1.0, 0.0\\])\n",
     "* **fov_degrees**: *float (optional)*\n",
     "    * the angle defining the extent of the 3D world that is seen from bottom to top of the camera view\n",
-    "    * Default: 50.0\n",
+    "    * Default: 75.0\n",
     "\n",
     "`AgentData` contains the following:\n",
     "* **times** : *np.ndarray (shape = [timesteps])*\n",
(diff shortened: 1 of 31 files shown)


schoinh (approved):
Great!

meganrm (approved):
This is the first time I'm looking at this repo in a while, but for all the python package converters, it seems like it would be possible to have a unified defaults file? and then have each converter possibly manipulate the data a little bit, ie to an list/np array. Just so you don't have to adjust values in so many places like this.

EricJIsaac (approved):

blairlyons:
> This is the first time I'm looking at this repo in a while, but for all the python package converters, it seems like it would be possible to have a unified defaults file? and then have each converter possibly manipulate the data a little bit, ie to an list/np array. Just so you don't have to adjust values in so many places like this.

good idea, although all but 2 references in this case are in documentation. I could probably make the documentation less repetitive

PR #31, opened 2021-06-25
Feature/cleanup and documentation update

blairlyons:
Problem
=======
Some code was inconsistent for doing the same things, and there was redundant documentation. 

For many of the supported simulation engines, the converter used to loop through the data twice, first to get the dimensions and create appropriately sized data structures, and then to read the data. This was less efficient and inconsistent between simulators.

Solution
========
I implemented a dynamic buffer for the converters, so that it initializes the data structures with a default size, and then if it runs out of space while reading, it creates larger arrays, incrementing by the default size.

I also removed the type_IDs array. It wasn't used in conversion for most simulators, and having both type IDs and type names was confusing. I removed the ReorderAgentsFilter that manipulated type IDs, we should implement features on the front end, like displaying the type mapping in the order it is provided, rather than having users set specific IDs for the types.

Also cleaned up jupyter notebooks and linked to documentation, and added camera defaults as constants.

I'm sorry this is a huge change set!

## Type of change

Change summary:
---------------
* use dynamic arrays for reading data from simulation engines
* only generate type IDs before writing to simularium file, use types list otherwise
* jupyter notebooks link to documentation instead of redundantly repeating it
* added camera defaults to constants
* renamed one character variables (t, n, s)

Keyfiles:
-----------------------
Documentation:
1. examples/Tutorial... - jupyter notebooks with links to documentation instead of redundant documentation

Constants:
1. simulariumio/constants.py - added constants for default buffer size and camera defaults
2. simulariumio/data_objects/camera_data.py - use constants
3. simulariumio/file_converter.py - use constants for camera defaults when updating file version

Dynamic Buffers:
1. simulariumio/data_objects/agent_data.py - add n_timesteps since buffer size might be longer than actual number of timesteps, add from_dimensions() class function to create empty AgentData with the given sized numpy arrays (buffers), add get_dimensions(), increase_buffer_size(), and check_increase_buffer_size() functions for dynamic buffers
2. simulariumio/data_objects/dimension_data.py - created DimensionData object to store buffer dimensions
3. simulariumio/cytosim/cytosim_converter.py - remove _parse_dimensions() functions, check if buffer needs to be increased in _parse_objects()
4. simulariumio/filters/every_nth_timestep_filter.py - use new AgentData.from_dimensions() to create filtered data
5. simulariumio/mcell/mcell_converter.py - remove _count_agents_in_binary_cellblender_viz_frame() function, check if buffer needs to be increased in _read_binary_cellblender_viz_frame()
6. simulariumio/medyan/medyan_converter.py - remove _parse_data_dimensions(), check if buffer needs to be increased in _get_trajectory_data()
7. simulariumio/smoldyn/smoldyn_converter.py - remove _parse_dimensions() function, check if buffer needs to be increased in _parse_objects()
8. simulariumio/springsalad/springsalad_converter.py - remove _parse_dimensions() function, check if buffer needs to be increased in _parse_springsalad_data()
9. simulariumio/tests/converters/buffer/test_buffer_size.py - added tests for incrementing buffer size

Remove Type IDs:
1. simulariumio/filters/every_nth_agent_filter.py - use type names instead of type IDs
2. simulariumio/filters/every_nth_subpoint_filter.py - use type names instead of type IDs
3. simulariumio/filters/reorder_agents_filter.py - deleted, no longer works without type IDs, should be done differently anyway
4. simulariumio/filters/translate_filter.py - use type names instead of type IDs
5. simulariumio/data_objects/agent_data.py - remove type IDs
6. simulariumio/cytosim/cytosim_converter.py - don't set type IDs
7. simulariumio/physicell/physicell_converter.py - don't set type IDs
8. simulariumio/readdy/readdy_converter.py - use a local type_IDs variable for filtering but don't set type IDs in AgentData


simulariumio/trajectory_converter.py
```diff
@@ -122,19 +122,19 @@ def _read_trajectory_data(input_data: TrajectoryData) -> Dict[str, Any]:
...
             ] = TrajectoryConverter._get_spatial_bundle_data_subpoints(
-                input_data.agent_data
+                input_data.agent_data, type_ids
             )
         else:
             spatialData[
                 "bundleData"
             ] = TrajectoryConverter._get_spatial_bundle_data_no_subpoints(
-                input_data.agent_data
+                input_data.agent_data, type_ids
```
schoinh, line 137: Everywhere else you're removing type_ids but how come you're adding it here?

schoinh: Oh I think I got it... we still want to write type_ids to the output simularium file so this is doing that?

blairlyons: yeah this is basically the one place they are generated now. I realized I can also remove it in TrajectoryConverter.add_number_of_agents_plot() too

simulariumio/data_objects/agent_data.py
```diff
@@ -114,31 +114,33 @@ def __init__(
...
         bundle_data = buffer_data["spatialData"]["bundleData"]
         total_steps = len(bundle_data)
         max_n_agents = 0
         max_n_subpoints = 0
-        for t in range(total_steps):
-            data = bundle_data[t]["data"]
-            i = 0
+        for time_index in range(total_steps):
+            data = bundle_data[time_index]["data"]
+            buffer_index = 0
```
toloudis, line 132: just from reading this, it's hard to tell at a glance what buffer_index is indexing.  I think that `data` is just an array of numbers, that is a packed list of agents, and buffer_index keeps track of where we are in that array, right?

blairlyons: yes, maybe `data_index` instead? The data is packed in it's actual buffer form at this point, I realize I'm overusing "buffer" at this point

simulariumio/filters/add_agents_filter.py
```diff
@@ -33,5 +33,5 @@ def apply(self, data: TrajectoryData) -> TrajectoryData:
         Add the given agents to each frame of the simularium data
         """
         print("Filtering: add agents -------------")
-        data.agent_data.append_agents(self.new_agent_data)
+        data.agent_data = data.agent_data.append_agents(self.new_agent_data)
```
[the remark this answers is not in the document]
blairlyons: ok I fixed this, with the `append_agents()` function on `AgentData` I couldn't set the object itself to the new value, requiring it to be awkward and non-intuitive like this (or the tests fail). So I moved `append_agents()` to `TrajectoryData` because it can set it's `agent_data`, and also made the function names and comments clear about when an `AgentData` is getting copied. It doesn't need to be copied here because it should already be copied by `TrajectoryConverter` before it called `apply()` on this filter.

simulariumio/data_objects/agent_data.py
toloudis: i'm not sure how much better the name is going to get.  I suppose it might be easier to rename `data`-->`buffer`... But maybe a small comment saying that `data` is a list of numbers being the packed agent data, might be good enough.

schoinh (approved):

toloudis (approved):

blairlyons:
I ran some benchmarks and the results are that the dynamic buffers are less performant than the previous way of going through the data twice, by ~2-6x depending on the trajectory. It probably depends on the default buffer size too, but increasing the buffer is just a big performance hit. 

So I've updated the engine converters to go through the data and get the dimensions first again. I left `DimensionData` and most of the `AgentData` buffer functions because they're still useful for adding agents in a filter and creating empty `AgentData`.

I also had to change the ReaDDy dependency to use the prod release, the current dev release broke something. It's probably better for that to be on prod anyway, and I'm going to let the ReaDDy developers know.

And I added benchmarks to the repo, the cytosim input file requires git lfs since it's large.

blairlyons:
this will also fix the build failure on main

blairlyons:
To compare the new changes https://github.com/allen-cell-animated/simulariumio/compare/7f1eb9f..6b6c35e

Key files updated:

Converters:
restored `_parse_dimensions()` functions and use them in `_read()` to create the appropriate sized `AgentData`
- simulariumio/cytosim/cytosim_converter.py
- simulariumio/mcell/mcell_converter.py
- simulariumio/medyan/medyan_converter.py
- simulariumio/smoldyn/smoldyn_converter.py
- simulariumio/springsalad/springsalad_converter.py
just made the functions static so easier to test separately, these converters weren't changed much in this PR:
- simulariumio/physicell/physicell_converter.py 
-  simulariumio/readdy/readdy_converter.py 

Central stuff
- simulariumio/constants.py - removed unneeded constants
- simulariumio/tests/buffer/test_buffer_cytosim.py - removed this test for dynamic buffer in cytosim
- simulariumio/data_objects/dimension_data.py - fixed a bug in `add()`
- simulariumio/data_objects/agent_data.py - just some variable renaming and constants update

schoinh (approved):

toloudis (approved):
Looks good.  I have a couple of general comments about making this scale better and potentially faster.

Knowing what I know about the simularium format, you could probably do this in a streaming kind of way, to reduce memory footprint.  Write out each agent as it is processed, rather than all in one big blob.   There may or may not be some trajectory formats that make that impossible but when possible it could be a big savings.  I'm not sure if there are any major drawbacks to writing the file in that way -- maybe error handling but that's all I can think of right now.  anyway, chunking things up into smaller pieces should at least consume less memory during the translation.  

Secondly, we definitely need some more granular profiling to see where the time bottlenecks are.  As we discussed, is it in reading the original, converting data, or writing out?  (If they are sufficiently disentangled to even measure that?)

PR #51, opened 2021-08-13
Fix/benchmarks

### LINKED ISSUE #48: Benchmark data in cloud
## What needs to happen?
Upload benchmark data to Quilt and implement downloading it from there when running benchmarks.

## Why should we do this?
So other developers can run benchmarks without adding large benchmark files to the repository.

## When does this need to get done?
August 2021

blairlyons:
Problem
=======
Benchmark data was left out of the repo because the files were too large to commit
closes #48 

Solution
========
Added scripts to upload and download the benchmark data to/from Quilt and then use it to run benchmarks. The benchmarking setup could be nicer, but at least the data is in the cloud now. The AWS bucket I'm using isn't public yet, I don't have permissions to do that. So more work is needed to make this usable by others.

@JacksonMaxfield helped with example scripts from AICSImageIO, thanks!!

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

Change summary:
---------------
* upload benchmarking data to Quilt
* download benchmarking data from Quilt
* run benchmarks with downloaded data

Steps to Verify:
----------------
1. log in to sccb aws account
2. install in editable mode: 
```
cd simulariumio/
pip install -e .[dev]
```
2. run benchmarks/benchmark_simulariumio.py: 
```
cd benchmarks/
python benchmark_simulariumio.py
```

Keyfiles:
-----------------------
1. benchmarks/benchmark_simulariumio.py - run benchmarks, downloading data from Quilt if needed
2. benchmarks/download_benchmark_resources.py - download data from Quilt
3. benchmarks/upload_benchmark_resources.py - upload data to Quilt


EricJIsaac (approved):

benchmarks/benchmark_simulariumio.py
```diff
@@ -53,20 +62,31 @@ def convert_condensate_springsalad(input_path):
...
+        if os.path.isdir(os.path.join(LOCAL_RESOURCES_DIR, item)):
+            found_data = True
+            break
+    if not found_data:
+        args = Args()
+        download_benchmark_resources(args)
+    # run benchmarks
+    for item in os.listdir(LOCAL_RESOURCES_DIR):
+        item_path = os.path.join(LOCAL_RESOURCES_DIR, item)
+        if not os.path.isdir(item_path):
```
schoinh, line 86: Hm, how come we need to use `item_path` instead of just `item` now?

schoinh (approved):

benchmarks/benchmark_simulariumio.py
blairlyons: because the item isn't in the current working directory anymore, so we can't just use the filename, we have to include the path. Guess I could have prepended "resources/" here, but I put it all in a constant instead

benchmarks/TEST_RESOURCES_HASH.txt
```diff
@@ -0,0 +1 @@
+81cbf23054ce82abd06d511f04dd57b47a02b120552f3a3c770f3e78e8b6be9e
```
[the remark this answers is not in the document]
blairlyons: yeah, we eventually want anyone to be able to run the benchmarks, and they need this to download the test data from the Quilt bucket (although they can't yet because the bucket still needs to be made public)

PR #52, opened 2021-08-20
Feature/write binary

### LINKED ISSUE #30: Write binary Simularium files
## Use Case

When a user is generating a simularium file, which they will then drag/drop or load from a URL in the Simularium Viewer, it would be nice to output the binary version of simularium file format as well as JSON, that way the files can be smaller, and download and/or load faster.

## Solution

Use the simularium binary format documentation (https://docs.google.com/document/d/1Hdvw3XieGrTHd7swILzMXLY-dbu-auJVDvt9K1Zbb_Y/edit#heading=h.xk07om75xuhb) and add functionality to TrajectoryConverter to write_binary() and write_external_binary().

## Acceptance Criteria

A file generated from the binary writing functions can be loaded into the Simularium Viewer (once binary file reading is added).

blairlyons:
Problem
=======
SimulariumIO only outputs the JSON version of .simularium format, it would be nice to also output binary.
closes #30 

Solution
========
Split file writing functionality from `TrajectoryConverter` into a few writer objects, including abstract base `Writer`, `JSONWriter`, and `BinaryWriter`. Each of those implements `Writer.save()`, which replaces the old `TrajectoryConverter.write_external_JSON()`. The old `TrajectoryConverter.write_JSON()` functions are replaced with `TrajectoryConverter.save()` which uses the writers and defaults to writing binary (but can set `binary=False` to write JSON).

**Note** this changes the `TrajectoryConverter` API, renaming `write_JSON()` to `save()` and removing `write_external_JSON()`.

## Type of change

Change summary:
---------------
* move JSON writing to `Writer` and `JsonWriter`
* add `BinaryWriter` to write to binary format
* split binary files so they don't exceed size limits by frames or bytes
* update `TrajectoryConverter` functions
* add tests that check the float values in the buffer before writing binary are correct
* update tests and docs
* update jupyter notebook examples

Keyfiles:
-----------------------
1. simulariumio/constants.py - added constants for binary header, version, and end of frame symbols and max frames and bytes per file
2. simulariumio/trajectory_converter.py - moved `_read_trajectory_data()` to `JsonWriter.format_trajectory_data()` and moved part of `_get_spatial_bundle_data_subpoints()` to `Writer._get_frame_buffer()` so it can also be used by `BinaryWriter`. Updated `to_JSON()` to use `JsonWriter`, changed `write_JSON()` to `save(binary=True)`, and removed `write_external_JSON()` since that's now handled by `JsonWriter.save()`.
3. simulariumio/writers/binary_chunk.py - added this object to store info about a chunk of data that will be saved to binary, which allows the trajectory to be split into multiple files if there are too many frames or if the data is too large
4. simulariumio/writers/binary_writer.py - format data as a numpy array of floats and then use `np.array.tofile()` to write to file.
5. simulariumio/writers/json_writer.py - moved `TrajectoryData` reading and JSON writing from `TrajectoryConverter` here, except moved parts of the old `TrajectoryConverter._read_trajectory_data()` to `Writer` in order to reuse it for writing binary.
6. simulariumio/writers/writer.py - abstract base `Writer` that defines an API with `format_trajectory_data()` and `save()` and also has a few functions used by both implementing classes.

Discussion:
-----------------------
1. What should the binary file size limit be? It wasn't specified in the binary format spec, so I chose 500 MB for now, based on my experience loading JSON files in the Viewer, where ~200MB is the max I got to load in a reasonable time. Binary files seem to be about 25% the size of JSON (i.e. max of ~50MB), but I know they will import much faster and it will depend on where the bottlenecks are. So I just multiplied it 10x assuming that's close to the max the viewer could handle, but it's a pretty wild guess and might be way off.
2. A few of the values in the binary format spec are 4-byte ints (NF, F#, NA) and the rest of the numbers are 4-byte floats. I'm using numpy to write to file and I have to give it a dtype. Some of the ints are per frame, if there was a consistent frame size I could define dtype with mixed types, but since there isn't I think I can only choose one or the other? Or maybe there's a way to build a mask somehow? And I'm not exactly sure how the file type string works with this too, I'm using `float(ord(i))` to pack those as floats right now, not sure if that's correct. Anyway, it seems like it would be easier if we could have all the values be floats, unless someone knows a way to handle this? Guess I could use something other than numpy too, but the `array.tofile()` function is so convenient...
3. How do we handle trajectoryInfo and  plotData blocks?


toloudis:
https://docs.python.org/3/library/struct.html#struct.pack

PR #54, opened 2021-09-08
Feature/specify geometry

### LINKED ISSUE #41: Agent geometry
## Use Case
It would be nice if users could specify geometry as either mesh file or PDB file URLs for each agent and have it saved in the .simularium data.

## Acceptance Criteria
Users can provide filenames for each agent type while using any Converter and have it saved in the .simularium data.

## Details
.simularium file format v3 includes "geometry" in `data["trajectoryInfo"]["typeMapping"]` (https://docs.google.com/document/d/1hxRQ1quZecX3Lo7WEDLN-MPU1Ebiiv6f9WKMt8i_8ZE) but users of simulariumIO cannot set these.

Users should be able to set a `display_type` and `url` for each agent in the `typeMapping`, `display_type` options are `SPHERE` (default), `MESH`, `PDB`, `CUBE`, `GIZMO`.

The documentation should say that only the first 100,000 geometry files will be loaded by the Viewer.

This should be added to each Converter.

blairlyons:
Problem
=======
Users can't specify geometry information to be saved in .simularium files
closes #41 

Solution
========
Add a `DisplayData` object to collect geometry information from users. When writing the type mapping for a .simularium file, use this info to create the geometry block within the type mapping.

All the different converters had different ways of collecting per-type information, previously just display names and radii. Sometimes the output files have them and sometimes they don't, sometimes the user could override it and sometimes not. To make things more consistent, I moved name and radius to `DisplayData` and updated all the converters.

Note: this change is breaking imo since a trajectoryInfo version of 3 currently does not load in the viewer.

## Type of change

Change summary:
---------------
* Added `DisplayData` object to store geometry (display type, url, and color) as well as display names and radii
* When the type mapping is created in `AgentData`, use `DisplayData` to add a geometry section to the type mapping
* Update the converter for each engine so users can use `DisplayData` to specify geometry as well as display names and radii
* Add a check that whenever it is specified, display_type is consistent with viz_type and n_subpoints
* Make default parameters for collections None
* Update tests, docs, and examples

Keyfiles:
-----------------------
1. simulariumio/constants.py - added `DISPLAY_TYPE` with the display options
2. simulariumio/cytosim/cytosim_agent_info.py - removed this object and replaced it with a more general `DisplayData` object that other converters can use
3. simulariumio/cytosim/cytosim_converter.py - update to use new `DisplayData` and set `AgentData`'s `display_data`
4. simulariumio/data_objects/agent_data.py - add `display_data` to hold geometry info, add display data to type mapping when it's generated in `get_type_ids_and_mapping()`, add reading display data in `from_buffer_data()`
5. simulariumio/data_objects/display_data.py - added `DisplayData` object to hold names, radii, and geometry data
6. simulariumio/file_converter.py - update files read in to v3, all the new fields are optional so don't need to add any defaults
7. simulariumio/mcell/mcell_converter.py - update to use new `DisplayData` and set `AgentData.display_data`
8. simulariumio/mcell/mcell_data.py - replace `display_names` with `display_data `
9. simulariumio/medyan/medyan_agent_info.py - removed this object and replaced it with a more general `DisplayData` object that other converters can use
10. simulariumio/medyan/medyan_converter.py - update to use new `DisplayData` and set `AgentData.display_data`
11. simulariumio/medyan/medyan_data.py - replaced `agent_info` with `display_data` to store `DisplayData` for each agent, use `agents_with_endpoints` list to specify agents to draw endpoints for
12. simulariumio/physicell/physicell_converter.py - update to use new `DisplayData` and set `AgentData.display_data`
13. simulariumio/physicell/physicell_data.py - replace `types` mapping with `display_data` to store DisplayData, and `phase_names` to store cell phase names
14. simulariumio/readdy/readdy_converter.py - update to use new `DisplayData` and set `AgentData.display_data`
15. simulariumio/readdy/readdy_data.py - replace `radii` with `display_data`
16. simulariumio/smoldyn/smoldyn_converter.py - update to use new `DisplayData` and set `AgentData.display_data`
17. simulariumio/smoldyn/smoldyn_data.py - replace `radii` and `display_names` with `display_data`
18. simulariumio/springsalad/springsalad_converter.py - update to use new `DisplayData` and set `AgentData.display_data`
19. simulariumio/springsalad/springsalad_data.py - replace 'display_names` with `display_data`
20. simulariumio/trajectory_converter.py - add `_check_types_match_subpoints()` function and check before creating the buffer format, increment trajectory info version to write


simulariumio/data_objects/display_data.py
```diff
@@ -0,0 +1,89 @@
...
+        Parameters
+        ----------
+        name : str
+            A string display name for this type of agent
+            Default: use names from simulator data if possible
+        radius : float (optional)
+            A float radius for rendering this agent.
+            For fibers, this is the thickness of the line
+            For default agents, this is the scale of the representation
+            Default : 1.0
```
meganrm, line 42: why isn't the default set in `__init__` for this? Is it handled somewhere else?

simulariumio/data_objects/trajectory_data.py
```diff
@@ -29,9 +29,9 @@ def __init__(
         self,
         meta_data: MetaData,
         agent_data: AgentData,
-        time_units: UnitData = UnitData("s"),
-        spatial_units: UnitData = UnitData("m"),
-        plots: List[Dict[str, Any]] = [],
+        time_units: UnitData = None,
+        spatial_units: UnitData = None,
+        plots: List[Dict[str, Any]] = None,
```
meganrm, line 34: is this change unrelated to the display data?

simulariumio/data_objects/agent_data.py
```diff
@@ -178,6 +186,13 @@ def get_type_ids_and_mapping(self) -> Tuple[np.ndarray, Dict[str, Any]]:
                     last_tid += 1
                     type_id_mapping[tn] = tid
                     type_name_mapping[str(tid)] = {"name": tn}
+                    if (
+                        tn in self.display_data
+                        and not self.display_data[tn].is_default()
+                    ):
+                        type_name_mapping[str(tid)]["geometry"] = dict(
+                            self.display_data[tn]
+                        )
```
meganrm, line 195: I don't understand why you're initializing with one default, (SPHERE) but then ignoring the data if only has a default. It seems like you could either have no defaults, and then you just check if the object is empty, or you could just write out default data if none is give so we don't have to to deal with optional data.

simulariumio/data_objects/display_data.py
blairlyons: yeah some of the converters use it as an override only if it has more than a default value, otherwise they use the value in the data. If it isn't specified in the data or here, it does default to 1.0 in the converters. I could set it to 1.0 here but it's harder to do the float comparison than a None check.

simulariumio/data_objects/trajectory_data.py
blairlyons: yeah this is part of "make default parameters for collections None" to prevent future bugs, if they default to e.g. `[]` there can be weirdness when using them.

simulariumio/data_objects/agent_data.py
blairlyons: yeah, I started out without the "FIBER" display_type and this is leftover from that. I just updated it to default to None and only output the display_type if it's been set to something else.

simulariumio/tests/converters/test_cytosim_converter.py
```diff
@@ -322,38 +336,49 @@
...
+                            1: DisplayData(name="motor complex", radius=0.02)
+                        },
                         position_indices=[3, 4, 5],
                     ),
                 },
             ),
             {
                 "trajectoryInfo": {
-                    "version": 2,
+                    "version": 3,
```
meganrm, line 381: it would be good to make a CURRENT_VERSION constant held in the general tests folder that you import to all these files so you don't have to edit each one each time we bump the version

simulariumio/tests/converters/test_cytosim_converter.py
```diff
@@ -384,11 +409,26 @@
...
+                        "0": {
+                            "name": "fiber1",
+                        },
+                        "1": {
+                            "name": "actin",
+                            "geometry": {
+                                "displayType": "FIBER",
+                                "color": "#ffc100",
+                            },
+                        },
```
meganrm, line 421: with this whole expected result, it seems like it would be good to have the same expected simularium file for all your tests, so that again, there are less changes you need to make when you update the format.

simulariumio/tests/converters/test_cytosim_converter.py
meganrm: you could even put it in here: https://github.com/allen-cell-animated/simulariumio/blob/main/simulariumio/__init__.py

and have a get_file_format_version. Or you could just put it in constants.py. either way, you shouldn't be writing it in every one of these files

simulariumio/file_converter.py
```diff
@@ -82,5 +82,8 @@ def update_trajectory_info_version(self, data: Dict[str, Any]) -> Dict[str, Any]
                 "fovDegrees": DEFAULT_CAMERA_SETTINGS.FOV_DEGREES,
             }
             data["trajectoryInfo"]["version"] = 2
```
meganrm, line 84: shouldn't this also now be v = 3?

I think these (heere and line 87) should be `data["trajectoryInfo"]["version"] = self.current_trajectory_info_version`

blairlyons: No, this block updates 1 to 2, then lines 85-87 update 2 to 3. Guess I could combine them but this gives me a place to comment that there are no mandatory fields added, and keeps the structure consistent with transitions that will need defaults added.

meganrm: oh ok, yeah, you should definitely have a block that checks for v == 2, like you say for clarity. I just didn't read this carefully enough and was assuming the other checks were `if else`s. You could still have the separate check for v == 2, but still update v1 ==> v3 data, but it doesn't really matter either way.

simulariumio/file_converter.py
```diff
@@ -82,5 +82,8 @@ def update_trajectory_info_version(self, data: Dict[str, Any]) -> Dict[str, Any]
...
+        if int(data["trajectoryInfo"]["version"]) == 2:
+            # all the new fields from v2 to v3 are optional
+            data["trajectoryInfo"]["version"] = 3
         print(f"Updated TrajectoryInfo v1 -> v{self.current_trajectory_info_version}")
```
meganrm, line 88: this print statement should be in the v ==1 block.

blairlyons: No I just shouldn't hardcode v1, fixed now

simulariumio/data_objects/display_data.py
blairlyons: added a comment about this

simulariumio/tests/converters/test_cytosim_converter.py
blairlyons: yeah added it to constants, there's a separate version for each block

simulariumio/tests/converters/test_cytosim_converter.py
blairlyons: yeah moved it to conftest, only saves one repeat

simulariumio/constants.py
```diff
@@ -45,3 +45,18 @@ class DEFAULT_CAMERA_SETTINGS:
...
 DEFAULT_PLOT_MODE = "markers"
+
+
+class DISPLAY_TYPE:
+    SPHERE: str = "SPHERE"
+    CUBE: str = "CUBE"
+    GIZMO: str = "GIZMO"
+    PDB: str = "PDB"
+    OBJ: str = "OBJ"
+    FIBER: str = "FIBER"
```
toloudis, line 56: I would suggest removing CUBE and GIZMO at least until we actually support them (or e.g. comment them out and put a comment that says "coming up")

simulariumio/data_objects/agent_data.py
```diff
@@ -197,6 +212,27 @@ def get_type_names(
                 )
         return result
 
+    @staticmethod
+    def get_display_data(type_mapping: Dict[str, Any]) -> Dict[str, DisplayData]:
```
toloudis, line 216: is there a well defined type for the values in type_mapping?

simulariumio/data_objects/display_data.py
```diff
@@ -0,0 +1,94 @@
...
+
+log = logging.getLogger(__name__)
+
+###############################################################################
+
+
+class DisplayData:
+    name: str
+    radius: float
+    display_type: str
```
toloudis, line 16: instead of str, this should be DISPLAY_TYPE?

simulariumio/file_converter.py
```diff
@@ -44,6 +42,7 @@ def update_trajectory_info_version(self, data: Dict[str, Any]) -> Dict[str, Any]
         Update the trajectory info block
         to match the current version
         """
+        original_version = int(data["trajectoryInfo"]["version"])
         if int(data["trajectoryInfo"]["version"]) == 1:
```
toloudis, line 46: line 46 should use `original_version` now

simulariumio/file_converter.py
```diff
@@ -82,5 +81,11 @@ def update_trajectory_info_version(self, data: Dict[str, Any]) -> Dict[str, Any]
                 "fovDegrees": DEFAULT_CAMERA_SETTINGS.FOV_DEGREES,
             }
             data["trajectoryInfo"]["version"] = 2
-        print(f"Updated TrajectoryInfo v1 -> v{self.current_trajectory_info_version}")
+        if int(data["trajectoryInfo"]["version"]) == 2:
```
toloudis, line 84: is this still `original_version` here?

simulariumio/data_objects/agent_data.py
blairlyons: No, it's coming from a JSON file that was read back in. The keys are strings, but the values are strings, numbers, or another dict, depending on if each agent type has geometry info. I added to the function comment to clarify

simulariumio/data_objects/display_data.py
blairlyons: The fields in DISPLAY_TYPE are strings, so I was thinking if you do `DisplayData(display_type=DISPLAY_TYPE.SPHERE)` the DisplayData object you create has a string display_type = "SPHERE". I'm not as familiar with Python objects, I can make the type DISPLAY_TYPE? seems like then I'd need each display type to be a class deriving from DISPLAY_TYPE?

simulariumio/data_objects/display_data.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        Check if this DisplayData is only holding default data
+        """
+        return self.display_type is None and not self.url and not self.color
+
+    def to_string(self):
+        default = self.is_default()
+        return (
+            f"{self.display_type}: url={self.url}, color={self.color} "
+            f"DEFAULT? {default}"
```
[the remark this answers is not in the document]
blairlyons: I'm returning this string regardless of if_default(), I thought I couldn't embed it in the f-string itself, but I can so I updated it

simulariumio/file_converter.py
blairlyons: I can set it up differently, but these loops are not if else, so if the version starts at v1, it will get updated to v2 in the first loop, then v2 -> v3 in the next loop, etc. So it needs to keep checking the current version, not the cached original version. 

I'm only storing the original version here so I can report out which version it started from at the end. I changed it to just report that at the beginning of the function to be less confusing. I would get rid of the `original_version` variable completely but I can't embed `int(data["trajectoryInfo"]["version"])` in the f-string. I could also use `original_version` for the first loop logic, but unless I increment it in that loop, I'll need to keep checking the current version in the following `if` loops, not the original version.

simulariumio/file_converter.py
blairlyons: no because if it started as v1, the data should now be at v2 from going through the first loop and the `original_version` is still 1, no longer current. 

Megan was also confused by how I set up this function, but I didn't want to have to write a loop to go from v1 -> v2 and another to go from v1 -> v4 and have to redundantly redo the fields that are added between v1 and v2 in each loop. Is there a better way to write it though?

simulariumio/trajectory_converter.py
```diff
@@ -358,6 +361,43 @@ def _check_agent_ids_are_unique_per_frame(buffer_data: Dict[str, Any]) -> bool:
                 get_n_subpoints = True
         return True
 
+    @staticmethod
+    def _check_types_match_subpoints(trajectory_data: TrajectoryData) -> str:
```
toloudis, line 365: just curious, I wonder if this could be done per-agent rather than iterating over all the data.

simulariumio/data_objects/display_data.py
toloudis: I was picturing something like an Enum : https://docs.python.org/3/library/enum.html
It would be nice to be able to treat DISPLAY_TYPE as a type.  Something to think about anyway...
Also you might be able to do 
`display_type : DISPLAY_TYPE.SPHERE | DISPLAY_TYPE.FIBER...etc`
but 
`display_type: DISPLAY_TYPE` would be nicer

simulariumio/data_objects/display_data.py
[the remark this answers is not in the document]
toloudis: I was just thinking the string `DEFAULT?` didn't look very nice :)

simulariumio/file_converter.py
toloudis: i see, then my other suggestion is to stash `trajectory_info = data["trajectory_info"]` very early on.  It just gets to be a whole lot of string lookups where you could just do it once and it makes the code a little cleaner.

simulariumio/file_converter.py
meganrm: You could pull the converters out as functions. So the first if block could call, ‘convertv1tov2’ and convertv2tov3’, successively on the data.  And the second if else block would just call ‘convertv2tov3’

toloudis: i like that idea

simulariumio/data_objects/display_data.py
blairlyons: aha yeah I didn't realize I could do enums like that, it's updated now

simulariumio/file_converter.py
blairlyons: ok did that

blairlyons:
@toloudis and @meganrm here's a diff for the changes I made today https://github.com/allen-cell-animated/simulariumio/compare/6ff12e7...feature/specify-geometry, I think I addressed all your comments, lmk if not

simulariumio/cytosim/cytosim_converter.py
```diff
@@ -254,6 +255,20 @@ def _read(input_data: CytosimData) -> TrajectoryData:
...
+            for tid in input_data.object_info[object_type].display_data:
+                display_data = input_data.object_info[object_type].display_data[tid]
+                if "fiber" in object_type and (
+                    display_data.display_type != DISPLAY_TYPE.NONE
+                    and display_data.display_type != DISPLAY_TYPE.FIBER
+                ):
+                    display_data.display_type = DISPLAY_TYPE.FIBER
+                    print(
+                        f"{display_data.name} display type of "
+                        f"{display_data.display_type} was changed to FIBER"
```
toloudis, line 269: this needs to print the stringified version of the display_type.  (see if the enum can be modified to hold the string values maybe!)

simulariumio/data_objects/display_data.py
```diff
@@ -0,0 +1,126 @@
...
+        self.display_type = display_type
+        self.url = url
+        if color is not None and (
+            (len(color) != 4 and len(color) != 7) or color[0] != "#"
+        ):
+            raise DataError(f"{color} should be provided as '#xxxxxx' or '#xxx'")
+        self.color = color
+
+    @staticmethod
+    def display_type_to_string(display_type: DISPLAY_TYPE) -> str:
```
toloudis, line 78: i commented on the commit but this could be a member function of the enum class rather than living here.  alternatively you might be able to give the enum values as strings so as to avoid this function completely.  

but more importantly, a comment would be good just saying that these strings are permanent for file i/o reasons and any change to them must incur a version bump on the format.

simulariumio/file_converter.py
```diff
@@ -34,53 +32,78 @@ def __init__(self, input_path: str):
...
+            < CURRENT_VERSION.TRAJECTORY_INFO
         ):
             buffer_data = self.update_trajectory_info_version(buffer_data)
         self._data = TrajectoryData.from_buffer_data(buffer_data)
 
-    def update_trajectory_info_version(self, data: Dict[str, Any]) -> Dict[str, Any]:
+    @staticmethod
+    def update_trajectory_info_v1_to_v2(data: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Update the trajectory info block from v1 to v2
```
toloudis, line 43: add a docstring comment saying that this mutates the object passed in, and returns it.  it does not return a new copy

toloudis: also, why not pass in data["trajectoryInfo"] instead?  will these functions ever modify any other part of `data`?

simulariumio/medyan/medyan_converter.py
```diff
@@ -220,6 +239,20 @@ def _read(input_data: MedyanData) -> TrajectoryData:
...
+            for tid in input_data.display_data[object_type]:
+                display_data = input_data.display_data[object_type][tid]
+                if (
+                    display_data.display_type != DISPLAY_TYPE.NONE
+                    and display_data.display_type != DISPLAY_TYPE.FIBER
+                ):
+                    display_data.display_type = DISPLAY_TYPE.FIBER
+                    print(
+                        f"{display_data.name} display type of "
+                        f"{display_data.display_type} was changed to FIBER"
```
toloudis, line 253: again, make sure this is printing out the string form of the enum value

toloudis (approved):

simulariumio/data_objects/display_data.py
blairlyons: whoops didn't realize I could use strings as enum values, fixed that and added a comment

meganrm (approved):

EricJIsaac (approved):

PR #60, opened 2021-09-22
Feature/binary

### LINKED ISSUE #30: Write binary Simularium files

blairlyons:
Problem
=======
SimulariumIO only outputs the JSON version of .simularium format, it would be nice to also output binary.
closes #30 

Solution
========
Split file writing functionality from `TrajectoryConverter` into a few writer objects, including abstract base `Writer`, `JSONWriter`, and `BinaryWriter`. Each of those implements `Writer.save()`, which replaces the old `TrajectoryConverter.write_external_JSON()`. The old `TrajectoryConverter.write_JSON()` functions are replaced with `TrajectoryConverter.save()` which uses the writers and defaults to writing binary (but can set `binary=False` to write JSON).

**Note** this changes the `TrajectoryConverter` API, renaming `write_JSON()` to `save()` and removing `write_external_JSON()`. There's also a new module `writers`.

**Note** Production simularium.allencell.org needs updates that are currently on staging before this PR can be merged

## Type of change

Change summary:
---------------
* move JSON writing to `Writer` and `JsonWriter`
* add `BinaryWriter` to write to binary format
* split binary files so they don't exceed size limit by bytes
* update `TrajectoryConverter` functions
* add tests that check the float values in the buffer before writing binary are correct
* update tests and docs
* update jupyter notebook examples
* add spec to the file format for binary
* update binary writing to match new spec

Keyfiles:
-----------------------
I tried to comment on the diff too to make this easier to review.

1. file_format.md - updated spec to include binary files
2. simulariumio/__init__.py - added `writers` to main API
3. simulariumio/constants.py - added constants for binary format
4. simulariumio/trajectory_converter.py - moved `_read_trajectory_data()` to `JsonWriter.format_trajectory_data()` and moved part of `_get_spatial_bundle_data_subpoints()` to `Writer._get_frame_buffer()` so it can also be used by `BinaryWriter`. Updated `to_JSON()` to use `JsonWriter`, changed `write_JSON()` to `save(binary=True)`, and removed `write_external_JSON()` since that's now handled by `JsonWriter.save()`.
5. simulariumio/writers/binary_chunk.py - added this object to store info about a chunk of data that will be saved to binary, which allows the trajectory to be split into multiple files if the data is too large
6. simulariumio/writers/binary_values.py - added this object to store values to be saved in binary and their corresponding format strings
7. simulariumio/writers/binary_writer.py - format data and save as binary in `save()`
8. simulariumio/writers/json_writer.py - moved `TrajectoryData` reading and JSON writing from `TrajectoryConverter` here, except moved parts of the old `TrajectoryConverter._read_trajectory_data()` to `Writer` in order to reuse it for writing binary.
9. simulariumio/writers/writer.py - abstract base `Writer` that defines an API with `format_trajectory_data()` and `save()` and also has a few functions used by both implementing classes.


blairlyons:
from meeting early September:
- trajectoryInfo and plotData blocks are still in JSON only, can be saved to same file with binary spatialData
- to pack different types into binary: https://docs.python.org/3/library/struct.html#struct.pack (from Dan)

file_format.md
```diff
@@ -1,6 +1,10 @@
-# .simularium JSON Format
+# .simularium File Format
```
blairlyons, line 1: the file format is no longer just JSON, update descriptions of structure and files

file_format.md
```diff
@@ -37,10 +41,11 @@ JSON files accepted by the simularium-viewer contain the following data in JSON
       * visualization type
         * 1000 = default, rendered with PDB or mesh
         * 1001 = fiber, rendered as a line
+      * agent instance ID - integer number ID for this agent instance
```
blairlyons, line 44: whoops this was missing

file_format.md
```diff
@@ -37,10 +41,11 @@ JSON files accepted by the simularium-viewer contain the following data in JSON
...
       * agent type ID - integer number ID for the agent’s type. This is used to look up its display data (name and geometry)
       * position X, Y, Z - agent’s 3D position
       * rotation X, Y, Z - euler angles for agent’s orientation
-      * radius - the radius the agent occupies, used for drawing a sphere
+      * radius - the radius the agent occupies, used for drawing a sphere or for scaling other geometry representations
```
blairlyons, line 48: update for geometry

file_format.md
```diff
@@ -235,3 +243,55 @@ JSON files accepted by the simularium-viewer contain the following data in JSON
     }
 }
 ```
+
```
blairlyons, line 246: added spec for binary files, which we discussed here https://github.com/simularium/simularium-viewer/issues/123

simulariumio/__init__.py
```diff
@@ -30,3 +30,8 @@ def get_module_version():
     HistogramPlotData,
 )
 from .constants import DISPLAY_TYPE  # noqa: F401
+
```
blairlyons, line 33: make the writers accessible from the main module

simulariumio/constants.py
```diff
@@ -74,6 +75,58 @@ class CURRENT_VERSION:
 
 DEFAULT_BOX_SIZE = 100.0 * np.ones(3)
 
+
```
blairlyons, line 78: added constants for binary format

simulariumio/trajectory_converter.py
```diff
@@ -56,374 +53,6 @@ def __init__(self, input_data: TrajectoryData):
         """
         self._data = input_data
```
blairlyons, line 55: moved these functions to the writers, split up between JsonWriter and base Writer (for functions that are reused by BinaryWriter)

simulariumio/trajectory_converter.py
```diff
@@ -469,9 +98,12 @@ def add_number_of_agents_plot(
 
         Parameters
         ----------
-        agent_data: AgentData
-            The data shaped as an AgentData object
-            Default: None (use the currently loaded data)
+        plot_title: str
```
blairlyons, line 101: whoops this docstring was wrong

simulariumio/trajectory_converter.py
```diff
@@ -507,40 +139,35 @@ def to_JSON(self):
         Return the current simularium data in JSON format
 
         """
-        buffer_data = TrajectoryConverter._read_trajectory_data(self._data)
+        buffer_data = JsonWriter.format_trajectory_data(self._data)
```
blairlyons, line 142: moved this functionality to JsonWriter, I think this function could still be convenient

simulariumio/trajectory_converter.py
```diff
@@ -507,40 +139,35 @@ def to_JSON(self):
         Return the current simularium data in JSON format
 
         """
```
blairlyons, line 141: updated these functions since they've been moved to the writers. `TrajectoryConverter.write_external_JSON()` is now `JsonWriter.save()`

simulariumio/writers/binary_chunk.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
```
blairlyons, line 1: BinaryChunk is used to determine how many files to save and how to chunk up the data, if the data is larger than the max file size

simulariumio/writers/binary_values.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
```
blairlyons, line 1: `BinaryValues` stores values to be saved in binary and their corresponding format strings

simulariumio/writers/binary_writer.py
```diff
@@ -0,0 +1,417 @@
...
+        constant_first_frame_offset: int,
+        spatial_header_constant_size: int,
+        spatial_data_length: int,
+    ) -> BinaryValues:
+        """
+        Return spatial data header values and format
+        """
+        first_frame_offset = constant_first_frame_offset + chunk.n_frames
+        chunk.frame_offsets = [
+            offset + first_frame_offset for offset in chunk.frame_offsets
```
blairlyons, line 119: @toloudis should the frame offsets be the global offset from the start of the file, or the offset within the spatial data block? I assumed global would be easier for parsing, but wanted to check

toloudis: If this is what I think it is,  then I think it should be offsets within the spatial data block.  Mainly so that we aren't changing too many things. We already have code that expects these offsets to be that.  I'll be able to answer more definitively when I really dig into this code in a couple days.

blairlyons: ok let me know and I can make that change

toloudis: yes please

toloudis:
A couple of issues as I start writing a typescript parser for this.  
1.  The example file you gave me is little endian.  Make sure that the python code is enforcing/guaranteeing this.  
2. For the headers of individual blocks, the spec says that blocktype comes before block size but in the file you gave me, block size came first.
3. The block size saved into the block actually is the block offset, not its size. (This one renders the file unreadable till it's fixed)
4. Make sure (I haven't checked) that the block size includes the block header data (its size and type). We should note that in the spec.  To get the raw data from the block you have to read (blocksize-blockheadersize) bytes from (blockoffset+blockheadersize) offset.

toloudis:
One more thing I have noticed.  I think for the spatial data, it will be convenient (and possibly necessary) to have a byte length as a part of each frame in the spatial data info.  I think the back end needs to be able to transmit bundles of data that don't necessarily end on frame boundaries, and having that byte length would allow us to know where the frame was split. (Much like the EOF signature).

blairlyons:
@toloudis ok 1 is done.

For 2-4, my tests are checking those things, or I just checked them, and they should be correct. But I didn't do the offsets correctly, I forgot to go back and multiply to put them in bytes, seems like that would do it. Now all offsets and lengths are in bytes.

I also updated the offsets within the spatial block to be local to the block and added frame lengths. Just like for the blocks in the file header they are intermixed with frame offsets (i.e. frame 0 offset, frame 0 length, frame 1 offset, frame 1 length, ...).

Data is updated at same URL. Let me know if anything else is off!

simulariumio/data_objects/agent_data.py
```diff
@@ -429,12 +429,19 @@ def from_dimensions(
             ),
         )
 
+    def total_steps(self) -> int:
+        """
+        Get number of frames
+        Use n_timesteps to limit times if it has been provided
+        """
+        return self.n_timesteps if self.n_timesteps >= 0 else len(self.times)
+
```
blairlyons, line 438: this is convenience for `BinaryWriter`

simulariumio/file_converter.py
```diff
@@ -19,24 +21,207 @@
...
+            A InputFileData object containing .simularium data to load
         """
-        print("Reading Simularium JSON -------------")
-        buffer_data = json.loads(input_file.get_contents())
+        try:
+            print("Reading Simularium JSON -------------")
+            buffer_data = json.loads(input_file.get_contents())
+        except UnicodeDecodeError:
+            print("Reading Simularium binary -------------")
+            buffer_data = FileConverter._load_binary(input_file)
```
blairlyons, line 36: update to allow `FileConverter` to read in binary files

simulariumio/tests/conftest.py
```diff
@@ -566,3 +566,203 @@ def mixed_agents() -> TrajectoryData:
...
+    )
+    for frame_index in range(len(test_buffer["spatialData"]["bundleData"])):
+        test_frame = test_buffer["spatialData"]["bundleData"][frame_index]
+        expected_frame = expected_buffer["spatialData"]["bundleData"][frame_index]
+        assert test_frame["frameNumber"] == expected_frame["frameNumber"]
+        assert test_frame["time"] == expected_frame["time"]
+        assert False not in np.isclose(
+            np.array(test_frame["data"]), np.array(expected_frame["data"])
+        )
+    assert test_buffer["plotData"] == expected_buffer["plotData"]
```
blairlyons, line 768: this allows testing floats for isclose() rather than equality

simulariumio/constants.py
```diff
@@ -74,6 +75,57 @@ class CURRENT_VERSION:
...
+    PLOT_DATA_JSON = 2
+    SPATIAL_DATA_BINARY = 3
+    # TRAJ_INFO_BINARY = 4  # coming soon
+    # PLOT_DATA_BINARY = 5  # coming soon
+
+
+class BINARY_SETTINGS:
+    HEADER: str = "SIMULARIUMBINARY"
+    VERSION: int = 2
+    MAX_BYTES: int = 4000000000  # 4GB is max for one file
```
toloudis, line 95: technically this is (4*1024*1024*1024)-1 but your number is more conservative.

simulariumio/constants.py
```diff
@@ -74,6 +75,57 @@ class CURRENT_VERSION:
...
+    HEADER_N_VALUES_PER_BLOCK: int = 3  # block offsets, types, lengths
+    BLOCK_HEADER_N_VALUES: int = 2  # block type, block length
+    SPATIAL_BLOCK_HEADER_CONSTANT_N_VALUES: int = (
+        2  # spatial data version, number of frames
+    )
+    FRAME_HEADER_N_VALUES: int = 3  # frame number, time stamp, number of agents
+    BYTES_PER_VALUE: int = 4
+    BLOCK_OFFSET_BYTE_ALIGNMENT: int = 4
+
+    def HEADER_N_INT_VALUES() -> int:
```
toloudis, line 107: can this be defined as an expression rather than a function?

simulariumio/constants.py
```diff
@@ -74,6 +75,57 @@ class CURRENT_VERSION:
...
+        """
+        Get the number of int values stored in the header of binary files,
+        used for packing binary data.
+        """
+        return (
+            BINARY_SETTINGS.HEADER_CONSTANT_N_VALUES
+            + BINARY_SETTINGS.N_BLOCKS * BINARY_SETTINGS.HEADER_N_VALUES_PER_BLOCK
+        )
+
+    def DEFAULT_BLOCK_TYPES() -> List[int]:
```
toloudis, line 117: make this a literal expression rather than a function

toloudis: also the docstring says "get the number of" but it should say "get the list of".   I'm not sure why this is needed when you have the BINARY_BLOCK_TYPE enum as its own thing already.

simulariumio/data_objects/agent_data.py
toloudis: as a reader of this code, I think it would help if you had consistent naming: choose one of `timesteps`, `steps` or `frames` if possible.  All 3 are used in the above function and it's confusing.

simulariumio/file_converter.py
toloudis: The above try/except would probably be a lot more efficient if you could just read the first 16 bytes of the file and compare them with "SIMULARIUMBINARY" with a regular old if statement.  That check could be encapsulated as a little function in FileConverter, like `if is_binary_file(input_file):`

toloudis: is it `get_contents` that raises the UnicodeDecodeError ?

simulariumio/file_converter.py
```diff
@@ -19,24 +21,207 @@
...
         if (
             int(buffer_data["trajectoryInfo"]["version"])
             < CURRENT_VERSION.TRAJECTORY_INFO
         ):
-            buffer_data = self.update_trajectory_info_version(buffer_data)
+            buffer_data = FileConverter.update_trajectory_info_version(buffer_data)
         self._data = TrajectoryData.from_buffer_data(buffer_data)
 
+    @staticmethod
+    def _read_binary_file(input_file: InputFileData) -> Dict[str, Any]:
```
toloudis, line 45: something about `_read_binary_file` vs `_load_binary`,  one of the two needs renaming because they sound the same.  Thinking out loud:  Maybe this one could be called `_get_data_views_from_file`.  or _read_file_into_memory_buffers

simulariumio/file_converter.py
```diff
@@ -19,24 +21,207 @@
...
+        """
+        Read a .simularium binary file and return multiple views of the data
+        """
+        result = {}
+        with open(input_file.file_path, "rb") as open_binary_file:
+            result["byte_view"] = open_binary_file.read()
+            result["n_values"] = int(
```
toloudis, line 52: is this `n_values` actually used anywhere outside of here?  the file might contain encoded text data which does't really match up with BYTES_PER_VALUE.  Also why should this result be a dict rather than a namedTuple or some simple dataclass?

toloudis: also wondering if struct.unpack copies the data?

toloudis: see this as possible alternative to the use of struct.unpack in this case:  https://numpy.org/doc/stable/reference/generated/numpy.frombuffer.html

simulariumio/file_converter.py
```diff
@@ -19,24 +21,207 @@
...
+                f'<{result["n_values"]}f', result["byte_view"]
+            )
+        return result
+
+    @staticmethod
+    def _binary_header(binary_data: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Parse header of data from a .simularium binary file
+        """
+        pos = len(BINARY_SETTINGS.HEADER)
```
toloudis, line 68: a tiny bit more naming confusion: this function is called _binary_header and we are parsing this thing called BINARY_SETTINGS.HEADER and comparing it with a variable called binary_id.  I suggest the BINARY_SETTINGS.HEADER be renamed to BINARY_SETTINGS.FILE_IDENTIFIER or something like that.

simulariumio/file_converter.py
```diff
@@ -19,24 +21,207 @@
...
+            result["int_view"] = struct.unpack(
+                f'<{result["n_values"]}i', result["byte_view"]
+            )
+            result["float_view"] = struct.unpack(
+                f'<{result["n_values"]}f', result["byte_view"]
+            )
+        return result
+
+    @staticmethod
+    def _binary_header(binary_data: Dict[str, Any]) -> Dict[str, Any]:
```
toloudis, line 64: this function could use a comment or docstring that describes the binary format of the header.  I really think you could get more readable code by just parsing the header using 2 or 3 struck.unpack calls on reads from the raw file bytes, for example in the way it's described in this post: https://stackoverflow.com/questions/30630090/parsing-binary-data-into-separate-variables-in-python

simulariumio/file_converter.py
```diff
@@ -19,24 +21,207 @@
...
+                "Binary file parsing only implemented for "
+                f"version > 1 up to current version {BINARY_SETTINGS.VERSION}, "
+                f"found {binary_version}"
+            )
+        return {
+            "n_blocks": header_values[2],
+            "block_offsets": header_values[3::3],
+            "block_types": header_values[4::3],
+            "block_lengths": header_values[5::3],
+        }
```
toloudis, line 91: again i feel this could be a dataclass or namedtuple rather than a dict

simulariumio/file_converter.py
```diff
@@ -19,24 +21,207 @@
...
+        block_offset = int(
+            binary_header["block_offsets"][block_index]
+            / BINARY_SETTINGS.BYTES_PER_VALUE
+        )
+        block_type = binary_data["int_view"][block_offset]
+        if block_type != binary_header["block_types"][block_index]:
+            raise DataError(
+                "Block type is not consistent for block "
+                f"#{block_index} : {block_type} != "
+                + binary_header["block_types"][block_index]
```
toloudis, line 111: in these error messages it won't be clear which number came from the header and which came from the block

simulariumio/file_converter.py
```diff
@@ -19,24 +21,207 @@
...
+        Parse spatial data binary block from a .simularium binary file
+        """
+        block_offset = (
+            int(
+                binary_header["block_offsets"][block_index]
+                / BINARY_SETTINGS.BYTES_PER_VALUE
+            )
+            + BINARY_SETTINGS.BLOCK_HEADER_N_VALUES
+        )
+        spatial_data_version = binary_data["int_view"][block_offset]
```
toloudis, line 159: storing `binary_data["int_view"]` as some other variable name like `data_as_ints` (for example) would probably make this feel a bit more concise and readable.   because it shows up so many times

simulariumio/file_converter.py
```diff
@@ -19,24 +21,207 @@
...
+        current_frame_offset = block_offset + 2 + 2 * n_frames
+        frame_info = binary_data["int_view"][block_offset + 2 : current_frame_offset]
+        frame_lengths = frame_info[1::2]
+        result = {
+            "version": spatial_data_version,
+            "msgType": 1,
+            "bundleStart": 0,
+            "bundleSize": n_frames,
+            "bundleData": [],
+        }
```
toloudis, line 170: another dict that's really a dataclass or namedtuple

simulariumio/file_converter.py
```diff
@@ -19,24 +21,207 @@
...
+            elif block_type == 1:
+                result["trajectoryInfo"] = FileConverter._binary_block_json(
+                    block_index, binary_header, data
+                )
+            elif block_type == 2:
+                result["plotData"] = FileConverter._binary_block_json(
+                    block_index, binary_header, data
+                )
+            elif block_type == 3:
+                result["spatialData"] = FileConverter._binary_block_spatial_data(
```
toloudis, line 218: what happens if there is a block of type 0 and a block of type 3 in the same file?  should we do a check in here and error out if that happens?    or things like two plotdata blocks or two trajectoryinfo blocks will always overwrite the result object.  worth noting in a comment at the very least.

simulariumio/file_converter.py
```diff
@@ -19,24 +21,207 @@
 class FileConverter(TrajectoryConverter):
     def __init__(self, input_file: InputFileData):
         """
-        This object loads the data in .simularium JSON format
-        at the input file
+        This object loads data from the input file in .simularium format.
```
toloudis, line 24: Since you moved the writing code out into a `writers` section, why not move the reading code into a `readers` section?

simulariumio/writers/binary_chunk.py
toloudis: The above is a perfect line for a comment in the code , right where it says "Object to store size info for a binary file"

simulariumio/writers/binary_writer.py
```diff
@@ -0,0 +1,447 @@
...
+            return BINARY_SETTINGS.BLOCK_OFFSET_BYTE_ALIGNMENT - remainder
+        else:
+            return 0
+
+    @staticmethod
+    def _frame_buffers_n_values(trajectory_data: TrajectoryData) -> List[int]:
+        """
+        Get the number of values in the bundle data buffer for each frame
+        """
+        total_steps = trajectory_data.agent_data.total_steps()
```
toloudis, line 44: frames or steps?

simulariumio/writers/binary_writer.py
```diff
@@ -0,0 +1,447 @@
...
+            result.append(
+                Writer._get_frame_buffer_size(frame_index, trajectory_data.agent_data)
+            )
+        return result
+
+    @staticmethod
+    def _header_n_bytes() -> int:
+        """
+        Get length of binary header in bytes
+        return 64 as long as there are 3 blocks, so doesn't need padding
```
toloudis, line 56: is this comment about "return 64 as long as 3 blocks" even needed?

simulariumio/file_converter.py
blairlyons: added `InputFileData._is_binary_file()` which reads the header to check for binary

simulariumio/file_converter.py
blairlyons: renamed to `_binary_data_from_file()`

simulariumio/file_converter.py
blairlyons: I made a dataclass `BinaryFileData` and switched to using `numpy.frombuffer()` 🎉

simulariumio/readers/simularium_binary_reader.py
```diff
@@ -0,0 +1,212 @@
...
+            )
+        return result
+
+    @staticmethod
+    def _parse_binary_header(data_as_bytes: bytes) -> BinaryBlockInfo:
+        """
+        Parse header of data from a .simularium binary file
+        """
+        id_length = len(BINARY_SETTINGS.FILE_IDENTIFIER)
+        pos = id_length + 3 * BINARY_SETTINGS.BYTES_PER_VALUE
```
toloudis, line 45: Do you already have a symbolic constant for this `3` here?

simulariumio/readers/simularium_binary_reader.py
```diff
@@ -0,0 +1,212 @@
...
+        _, _, binary_version, n_blocks = struct.unpack(
+            f"{id_length}sIII", data_as_bytes[:pos]
+        )
+        if binary_version != BINARY_SETTINGS.VERSION:
+            raise DataError(
+                "Binary file parsing only implemented for "
+                f"version > 1 up to current version {BINARY_SETTINGS.VERSION}, "
+                f"found {binary_version}"
+            )
+        block_info_length = 3 * BINARY_SETTINGS.BYTES_PER_VALUE * n_blocks
```
toloudis, line 55: Do you already have a symbolic constant for this 3 here?

simulariumio/readers/simularium_binary_reader.py
```diff
@@ -0,0 +1,212 @@
...
+        result = {}
+        binary_data = SimulariumBinaryReader._binary_data_from_file(input_file)
+        block_info = SimulariumBinaryReader._parse_binary_header(binary_data.byte_view)
+        # parse blocks
+        found_blocks = []
+        for block_index in range(block_info.n_blocks):
+            block_type_id = SimulariumBinaryReader._binary_block_type(
+                block_index, block_info, binary_data.int_view
+            )
+            if block_type_id == 0:
```
toloudis, line 177: isn't there a symbolic enum for these integer values?

simulariumio/readers/simularium_binary_reader.py
```diff
@@ -0,0 +1,212 @@
...
+                    f"WARNING: More than one {block_type} block found, "
+                    "only using last one"
+                )
+            if data_type == "JSON":
+                result[block_type] = SimulariumBinaryReader._binary_block_json(
+                    block_index, block_info, binary_data.byte_view
+                )
+            elif block_type == "spatialData":
+                result[
+                    "spatialData"
```
toloudis, line 200: maybe change this to `block_type` to match line 195?

simulariumio/writers/binary_chunk.py
```diff
@@ -0,0 +1,56 @@
...
+    n_values: int
+    first_frame_index: int
+    frame_n_values: List[int]
+
+    def __init__(
+        self,
+        first_frame_index: int = 0,
+    ):
+        """
+        Object to store size info for a binary file
```
toloudis, line 26:
```suggestion
        Object to store size info for a binary file.  BinaryChunk is used to determine how many files to save and how to chunk up the data, if the data is larger than the max file size.
```

simulariumio/writers/binary_writer.py
```diff
@@ -0,0 +1,441 @@
...
+            return BINARY_SETTINGS.BLOCK_OFFSET_BYTE_ALIGNMENT - remainder
+        else:
+            return 0
+
+    @staticmethod
+    def _frame_buffers_n_values(trajectory_data: TrajectoryData) -> List[int]:
+        """
+        Get the number of values in the bundle data buffer for each frame
+        """
+        total_steps = trajectory_data.agent_data.total_timesteps()
```
toloudis, line 44: If not changing this, then I think there needs to be some documentation somewhere that `frame` is synonymous with `timestep`, `step` and/or `time`.   When the code alternates between these terms interchangeably, I'd be worried that an engineer might be confused by lines like 
```
for frame_index in range(total_steps):
```
when it could easily be
```
for step_index in range(total_steps):
```

simulariumio/writers/binary_writer.py
```diff
@@ -0,0 +1,441 @@
...
+            max_bytes - header_n_bytes - traj_info_n_bytes - plot_data_n_bytes
+        )
+        file_chunks = [BinaryChunk()]
+        current_chunk = 0
+        current_frames_bytes = 0
+        for frame_index, buffer_size in enumerate(frame_buffers_n_values):
+            spatial_header_n_values = (
+                BINARY_SETTINGS.BLOCK_HEADER_N_VALUES
+                + BINARY_SETTINGS.SPATIAL_BLOCK_HEADER_CONSTANT_N_VALUES
+                + 2 * (file_chunks[current_chunk].n_frames + 1)
```
toloudis, line 128: is there already a symbolic constant for this `2`?

simulariumio/writers/binary_writer.py
```diff
@@ -0,0 +1,441 @@
...
+                + [
+                    block_offsets[0],
+                    block_types[0],
+                    block_n_bytes[0],
+                    block_offsets[1],
+                    block_types[1],
+                    block_n_bytes[1],
+                    block_offsets[2],
+                    block_types[2],
+                    block_n_bytes[2],
```
toloudis, line 192: Just for fun, I googled "interleave multiple lists in python" and found this:
```
[val for tup in zip(block_offsets, block_types, block_n_bytes) for val in tup]
```
assuming the lists are all the same length... which would be future-proof for arbitrary numbers of blocks.

simulariumio/writers/binary_writer.py
blairlyons: I did change it `total_steps()` to `total_timesteps`. I could go through the codebase and remove `frame`, but I know it's used a bunch of places

simulariumio/writers/binary_writer.py
blairlyons: no but I'll add one

simulariumio/writers/binary_writer.py
blairlyons: cool!

simulariumio/writers/binary_writer.py
toloudis: can be separate PR/issue but I think it deserves consistency.  Also see this thread for how another team discussed what the data of a timestep is: https://github.com/MDAnalysis/mdanalysis/issues/250#issuecomment-98279984

Issue #55, opened 2021-09-09
Support more Python versions


blairlyons:
## Use Case
Based on the Python features SimulariumIO uses, it will likely work on more versions than just 3.7 and 3.8. It would be nice if we could expand the list of supported versions so users working in another version know that they can use the package with their current Python version.

## Acceptance Criteria
Any Python versions that SimulariumIO works with should be tested and listed on the readme

## Details
Test each version once and decide which ones to add to the build action.

PR #62, opened 2021-10-01
Admin/update documentation

### LINKED ISSUE #56: Document Jupyter dependency for examples
## Use Case
For new users using the tutorial, it would be nice to make it more clear that they additionally need Jupyter to run the examples and give them a setup option to install with it in one step.

## Acceptance Criteria
There should be a "tutorial" requirements block in setup.py and more description in the readme of how to use it or install Jupyter separately in order to use the examples.

### LINKED ISSUE #61: Update documentation
## What needs to happen?
The readme should be updated to address issues found during paper revisions.

## Why should we do this?
To help users access SimulariumIO features
- list visualization options
- clarify sharing instructions

## When does this need to get done?
mid-Oct 2021

blairlyons:
Problem
=======
Readme was out of date and didn't provide enough detail, and there was no easy way to install the tutorial dependencies.
closes #56 
closes #61 

Solution
========
Updated readme and added a "tutorial" dependency group to setup.py

## Type of change

Change summary:
---------------
* Add to readme overall description to include the main user flow (convert, visualize, share)
* Add visualization capabilities to readme
* Update instructions for install, convert, visualize, and share on readme
* Add "tutorial" dependency group to setup.py that includes jupyter and any optional dependencies required to run the example notebooks (with exception of ReaDDy, since it uses conda and special install instructions for that are already on the readme)
* Add scipy dependency for MCell (was overlooked previously)
* Make data a required argument for `TrajectoryConverter.add_plot()` (was overlooked previously)

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,15 @@ Upload your file to a public Dropbox or Google Drive folder or an AWS S3 bucket,
 
 We're discussing the possibility of adding the ability to export Simularium files directly with the authors of some packages.
 
+Current rendering capabilities of the viewer include:
+* Surface topologies: users can provide mesh files or use default spheres
+* Protein Databank files: users can provide PDB files to render atoms in a molecule
+* Line rendering for fibers, filaments, or bonds
+
+Rendering capabilities planned for future:
+* Volume rendering for RDME or PDE-based simulation results
+
+
 <br/>
 
 ---
@@ -196,6 +205,25 @@ converter.add_plot(
 converter.save("output_file_name")
 ```
 
+### Render with meshes or PDB files
+If you'd like to render agents with arbitrary meshes instead of spheres, add a `DisplayData` for each agent type you'd like to be rendered with the mesh, and specify an OBJ mesh file. Alternately, you can provide a Protein Databank (PDB) file or PDB code to render atoms for a molecule.
+```python
+display_data={
+    "A" : DisplayData(
+        name="Molecule A",
+        display_type=DISPLAY_TYPE.PDB,
+        url="https://files.rcsb.org/download/3KIN.pdb",
+        color="#0080ff",
+    ),
+    "B" : DisplayData(
+        name="Molecule B",
+        display_type=DISPLAY_TYPE.OBJ,
+        url="molecule_b.obj",
+        color="#333333",
+    ),
+}
+```
+
 <br/>
 
 ---
@@ -204,7 +232,7 @@ converter.save("output_file_name")
 
 ## Visualize results
 1. In a supported browser (Firefox, Chrome, or Edge), navigate to https://simularium.allencell.org/viewer.
-2. Drag the file output from SimulariumIO from your file browser onto the window or use the file upload dialogue to choose your file
+2. Drag the file output from SimulariumIO from your file browser onto the window or choose Load model > From your device, and select your file from the file upload dialogue.
 
 <br/>
 
@@ -212,9 +240,11 @@ converter.save("output_file_name")
 
 <br/>
 
-## Share URL to results
+## Share link to results
 1. Upload your Simularium file to one of the supported public cloud providers, including Dropbox, Google Drive, or Amazon S3, and get a publicly accessible link to the file.
-2. In a supported browser, navigate to https://simularium.allencell.org/viewer?trajUrl= [link to your file]. You can share this link with collaborators or post it on your website so that others can interactively view your results.
+2. In a supported browser (Firefox, Chrome, or Edge), navigate to https://simularium.allencell.org.
+3. Choose Load model > From a URL. In the dialog, provide the URL to your .simularium file and choose Load. 
+4. Once the file is loaded, you can copy the page URL and share this link with collaborators or post it on your website so that others can interactively view your results.
 
 <br/>
 
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ converter.add_plot(
         },
     )
 )
-converter.save("output_file_name")
+converter.write_JSON("output_file_name")
 ```
 
 <br/>
@@ -180,7 +180,7 @@ input_data = TrajectoryData(
         rotations=ROTATIONS,
     )
 )
-TrajectoryConverter(input_data).save("output_file_name")
+TrajectoryConverter(input_data).write_JSON("output_file_name")
 ```
 
 ### Add metrics data to plot
@@ -202,7 +202,7 @@ converter.add_plot(
         },
     )
 )
-converter.save("output_file_name")
+converter.write_JSON("output_file_name")
 ```
 
 ### Render with meshes or PDB files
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -5,18 +5,79 @@ This repository is part of the Simularium project ([simularium.allencell.org](ht
...
+
+# 2. Load in the [Simularium Viewer](https://simularium.allencell.org/viewer)
+Load your file in the Simularium Viewer at https://simularium.allencell.org/viewer to interactively rotate and play back your data:
+
+![Loading a file in the Simularium Viewer](/examples/img/drag_drop.gif)
+
+<br/>
+
+# 3. Share by URL 
+Upload your file to a public Dropbox or Google Drive folder or an AWS S3 bucket, and use the URL https://simularium.allencell.org/viewer?trajUrl= [link to your file]
```
meganrm, line 71: for clarity, 
```suggestion
Upload your file to a public Dropbox or Google Drive folder or an AWS S3 bucket, and use the URL https://simularium.allencell.org/viewer?trajUrl=[LINK_TO_YOUR_FILE]
```
And maybe have an example that works (a link to one of our trajectories)

README.md
```diff
@@ -30,19 +91,44 @@ Simulariumio converts simulation outputs to the format consumed by the [Simulari
...
+
+* If ReaDDy trajectories will be converted, the ReaDDy python package must be installed:
+(add conda forge channel if it's not already: `conda config --add channels conda-forge`)
+`conda install -c readdy readdy`
+
+**Stable Release:** `pip install simulariumio`
+
+**Development Head:** `pip install git+ssh://git@github.com/allen-cell-animated/simulariumio.git`
+
+**Please note** that to run the Jupyter notebook examples you should also install Jupyter, either with `pip install jupyter`, or by installing SimulariumIO with the tutorial requirements: `pip install simulariumio[tutorial]`
```
meganrm, line 124: Nice 👍🏻

README.md
```diff
@@ -94,41 +186,82 @@ TrajectoryConverter(input_data).write_JSON("output_file_name")
...
-converter = TrajectoryConverter(input_data) # see above to create converter
-converter.add_plot(example_scatter_plot, "scatter")
 converter.write_JSON("output_file_name")
 ```
-___
 
-## Installation
-**Install Requires:** 
+### Render with meshes or PDB files
+If you'd like to render agents with arbitrary meshes instead of spheres, add a `DisplayData` for each agent type you'd like to be rendered with the mesh, and specify an OBJ mesh file. Alternately, you can provide a Protein Databank (PDB) file or PDB code to render atoms for a molecule.
```
meganrm, line 209: 1. "arbitrary meshes" seems like strange wording to me. Maybe "render agents as meshes"? They aren't arbitrary from the users point of view. 
2. I don't think we support just a PDB code? I could be wrong (@toloudis) but I think we still need a link or path to where that pdb file is. Or at least that's the only code I've looked at, maybe there is something that looks at the PDB first.

README.md
```diff
@@ -94,41 +186,82 @@ TrajectoryConverter(input_data).write_JSON("output_file_name")
...
+---
 
-**Development Head:** `pip install git+https://github.com/allen-cell-animated/simulariumio.git`
+<br/>
 
-___
+## Visualize results
+1. In a supported browser (Firefox, Chrome, or Edge), navigate to https://simularium.allencell.org/viewer.
+2. Drag the file output from SimulariumIO from your file browser onto the window or choose Load model > From your device, and select your file from the file upload dialogue.
+
```
meganrm, line 238: lines 235 - 249 seem redundant with what starts on line 25. ## Convert simulation outputs so they can be visualized in the [Simularium Viewer](https://simularium.allencell.org/).

Or am I missing something?

README.md
blairlyons: when I don't leave a space between `=[` the opening bracket gets highlighted as part of the URL, do you know how to escape it? I tried prepending `\` but that didn't work. maybe it doesn't matter

meganrm: I think that's fine, I would probably even remove the brackets so people don't think you need them. I know that's common in code, bu in this case it's just a string

README.md
blairlyons: 1. yeah the reviewer used that term and I got stuck on it, will update
2. would be great if @toloudis can confirm one way or another and I can update

README.md
blairlyons: That first section is meant to be a quick overview and this is more detail. But lmk if you have ideas to make that more clear

meganrm: is it possible to link down here from up there? Like have a "details" link for each of those overviews? Not a big deal either way.

meganrm (approved):

README.md
```diff
@@ -30,19 +91,45 @@ Simulariumio converts simulation outputs to the format consumed by the [Simulari
 * Conversions for data from custom engines can be implemented using the TrajectoryConverter class
 * Also accepts metrics data for plots to display alongside spatial data
 
-We're working to improve performance for converting large trajectories, and also discussing with the authors of some packages the possibility of adding the ability to export Simularium files directly.
+We're discussing the possibility of adding the ability to export Simularium files directly with the authors of some packages.
 
-___
+Current rendering capabilities of the viewer:
+* Spheres: by default, each agent in a scene is represented as a single sphere
+* Mesh surfaces: represent each agent as a mesh files, e.g. coarse molecular surfaces
```
toloudis, line 98:
```suggestion
* Mesh surfaces: represent each agent as a mesh (loaded from an .obj file), e.g. coarse molecular surfaces
```

README.md
```diff
@@ -30,19 +91,45 @@ Simulariumio converts simulation outputs to the format consumed by the [Simulari
...
-2. In a supported browser, navigate to https://simularium.allencell.org/viewer?trajUrl=[link to your file]. You can share this link with collaborators or post it on your website so that others can interactively view your results.
+---
 
-___
+<br/>
+
+## Installation
+**Install Requires:** 
+
+* Requires Python 3.7 or 3.8
```
toloudis, line 115: just curious, is 3.9 supported?

README.md
toloudis: Currently it has to be a full URL to a single .pdb file.

toloudis: in this case "arbitrary" means "any mesh of your choice" and not "a random mesh that is unrelated to anything".  It's arbitrary from the point of view of the software. :)

meganrm: yeah, I understand that, but it seems weird here since this is a doc that is supposed to be more user facing (I'm assuming).

toloudis: I agree with you.  I can see how users might interpret it that way.

README.md
blairlyons: I think it would work but I need to test, there's an issue for it #55

README.md
blairlyons: yeah I'll add some anchors https://gist.github.com/asabaylus/3071099

PR #75, opened 2021-11-02
Fix/require display type

### LINKED ISSUE #71: Make displayType required
## Use Case
It would be more clear for users and simplify code in the viewer if `displayType` was added for all agent types in SImulariumIO.

## Acceptance Criteria
All .simularium files have `geometry` with `displayType` for each agent type and the file format documentation reflects that these are required.

## Details
`geometry` and `displayData` blocks were previously optional, update the file format to specify they are required and add them to all trajectories, default values if nothing was specified by the user.

Also add notes to the file format about which fields will throw an error vs warning if missing.

Also specify that all geometry urls start with "http" unless they are PDB IDs (or add a field for the IDs?)

Should there also be a display type for CIF vs PDB?

blairlyons:
Problem
=======
Simulariumio only added "displayType" to .simularium files when something non-default was specified by the user
closes #71 

Solution
========
I modified `AgentData` so "displayType" is always added and defaults to either "SPHERE" or "FIBER" based on the number of subpoints the agent has.

## Type of change

Change summary:
---------------
* make sure any type in the mapping has geometry with displayType
* default to SPHERE or FIBER based on n_subpoints
* fix tests

Keyfiles:
-----------------------
1. simulariumio/data_objects/agent_data.py - add a `geometry` block with `displayType` in `get_type_ids_and_mapping()` if it isn't already
2. simulariumio/data_objects/display_data.py - add `check_set_default_display_type()` to set the displayType to the correct default if needed, always add "displayType" key to dict version

### DIFF
--- a/simulariumio/data_objects/agent_data.py
+++ b/simulariumio/data_objects/agent_data.py
@@ -14,6 +14,7 @@ from ..constants import (
     V1_SPATIAL_BUFFER_STRUCT,
     VIZ_TYPE,
     BUFFER_SIZE_INC,
+    DISPLAY_TYPE,
 )
 from .dimension_data import DimensionData
 from .display_data import DisplayData
@@ -178,22 +179,34 @@ class AgentData:
         last_tid = 0
         for time_index in range(total_steps):
             for agent_index in range(len(self.types[time_index])):
-                tn = self.types[time_index][agent_index]
-                if len(tn) == 0:
+                type_name = self.types[time_index][agent_index]
+                if len(type_name) == 0:
                     continue
-                if tn not in type_id_mapping:
+                if type_name not in type_id_mapping:
                     tid = last_tid
                     last_tid += 1
-                    type_id_mapping[tn] = tid
-                    type_name_mapping[str(tid)] = {"name": tn}
+                    type_id_mapping[type_name] = tid
+                    type_name_mapping[str(tid)] = {"name": type_name}
+                    has_subpoints = self.n_subpoints[time_index][agent_index] > 0
                     if (
-                        tn in self.display_data
-                        and not self.display_data[tn].is_default()
+                        type_name in self.display_data
+                        and not self.display_data[type_name].is_default()
                     ):
-                        type_name_mapping[str(tid)]["geometry"] = dict(
-                            self.display_data[tn]
+                        self.display_data[type_name].check_set_default_display_type(
+                            has_subpoints
                         )
-                type_ids[time_index][agent_index] = type_id_mapping[tn]
+                        type_name_mapping[str(tid)]["geometry"] = dict(
+                            self.display_data[type_name]
+                        )
+                    else:
+                        type_name_mapping[str(tid)]["geometry"] = {
+                            "displayType": (
+                                DISPLAY_TYPE.FIBER
+                                if has_subpoints
+                                else DISPLAY_TYPE.SPHERE
+                            ).name,
+                        }
+                type_ids[time_index][agent_index] = type_id_mapping[type_name]
         return type_ids, type_name_mapping
 
     @staticmethod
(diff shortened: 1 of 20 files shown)


toloudis (approved):

toloudis (approved):
color is next!! 😉

simulariumio/tests/conftest.py
```diff
@@ -22,8 +22,18 @@
...
+        "0": {
+            "name": "C",
+            "geometry": {
+                "displayType": "SPHERE",
+            },
+        },
+        "1": {
+            "name": "U",
+            "geometry": {
+                "displayType": "SPHERE",
```
meganrm, line 34: is there a reason on the types in the tests are set to defaults? Why not "PDB" for example?

simulariumio/tests/converters/test_springsalad_converter.py
```diff
@@ -86,12 +86,28 @@
...
+                            "name": "CYAN",
+                            "geometry": {
+                                "displayType": "SPHERE",
+                            },
+                        },
+                        "4": {
+                            "name": "BLUE",
+                            "geometry": {
+                                "displayType": "SPHERE",
+                            },
```
meganrm, line 109: I think someone has commented to this effect on previous PRs: but it seems like it would be possible to have a base test object, that you import into all these different test files, and maybe make minor edits to for the specific engine, but this seems like an excessive amount of very similar changes to a lot of files. If you don't want to do it for this PR, it would be good to write up an issue that is reducing repeated code across test files

simulariumio/tests/conftest.py
meganrm: OH! this is the expected resulting data, and is your test data never specifying the displayType? In any case, do your tests cover all cases? Not specified, and user specified?

simulariumio/data_objects/agent_data.py
```diff
@@ -178,22 +179,34 @@ def get_type_ids_and_mapping(self) -> Tuple[np.ndarray, Dict[str, Any]]:
...
                     last_tid += 1
-                    type_id_mapping[tn] = tid
-                    type_name_mapping[str(tid)] = {"name": tn}
+                    type_id_mapping[type_name] = tid
+                    type_name_mapping[str(tid)] = {"name": type_name}
+                    has_subpoints = self.n_subpoints[time_index][agent_index] > 0
                     if (
-                        tn in self.display_data
-                        and not self.display_data[tn].is_default()
+                        type_name in self.display_data
```
schoinh, line 192: I kinda struggled to understand this part of the code because I kept thinking `display_data` was of type `DisplayData` when it's actually a mapping of agent names to their `DisplayData`. I know this isn't new to this PR but I think it'd be helpful to rename `this.display_data` as `this.display_data_map`.

simulariumio/tests/conftest.py
blairlyons: yes, some of these are "user" specified and some are defaults, for this one I'm providing some stuff for C, W, S, Y, and L and others are defaults

simulariumio/tests/converters/test_springsalad_converter.py
blairlyons: the tox test runner does some weird caching that I don't totally understand, so for identical objects I've tried to do that, but sometimes have run into issues modifying a shared object and then seeing that modification in the next test that uses the hypothetically plain unmodified object. We could add an issue but it's not highest priority imo

simulariumio/data_objects/agent_data.py
```diff
@@ -178,22 +179,34 @@ def get_type_ids_and_mapping(self) -> Tuple[np.ndarray, Dict[str, Any]]:
...
+                        and not self.display_data[type_name].is_default()
```
schoinh, line 193: But even with the above realization I'm confused about this if-else branching... Could you please explain for me what kind of scenario the `if` condition deals with and what it does?? 🙏🏻

EricJIsaac (approved):

simulariumio/data_objects/agent_data.py
blairlyons: This is checking whether the display data for this type exists in the mapping and is not just a default display data, i.e. it was provided by the user or engine. If so, we want to use it (and make sure its display type is set), if not, we add a default one so the agent will have a display type

simulariumio/data_objects/agent_data.py
blairlyons: It probably would be a little clearer, but it would make this a breaking change and I'd prefer not to do that

PR #94, opened 2022-02-02
Fix/v3 updates

blairlyons:
Problem
=======
I noticed some bugs using v3 updates, and the file format spec is out of date, and also wanted to add the ability to skip frames in PhysiCell reading, like in MCell (this should be a separate PR but I did it at the same time while working with PhysiCell data, sorry!)

Solution
========
fixed bugs, added `nth_timestep_to_read` to `PhysicellData`, updated file format spec.

## Type of change
- [ ] New feature (non-breaking change which adds functionality)

Change summary:
---------------
* fixed errors if a geometry block was in a .simularium file being read back in with FileConverter
* added the ability to skip frames in PhysiCell reading
* updated file format spec
* added and updated tests
* updated PhysiCell example notebook

Keyfiles:
-----------------------
1. simulariumio/data_objects/agent_data.py - provide a name to `DisplayData` constructor to fix bug
2. simulariumio/data_objects/display_data.py - try to create a `DISPLAY_TYPE` enum from a string display type, since that's how it is saved in JSON, to fix bug. Also added more descriptive error messages
3. simulariumio/physicell/physicell_data.py - added `nth_timestep_to_read` like in `McellData`, to be able to skip timesteps when there's too much data.
4. simulariumio/physicell/physicell_converter.py - use `nth_timestep_to_read`
5. file_format.md - add version 3 updates


file_format.md
```diff
@@ -68,7 +68,17 @@ JSON files accepted by the simularium-viewer contain the following data in JSON
...
+        // model metadata
+        "trajectoryTitle" : "Fast diffusion",
+        "modelInfo" : {
+            "title" : "SARS-CoV-2 Dynamics in Human Lung Epithelium"
+            "version" : 4.1
+            "authors" : "Michael Getz et al"
+            "description" : "A PhysiCell model of SARS-CoV-2 dynamics in human lung epithelium."
+            "doi" : "10.1101/2020.04.02.019075"
+            "sourceCodeUrl" : "https://github.com/pc4covid19/pc4covid19"
+        },
```
schoinh, line 81: Should this and the geometry fields be updated in the above bulleted list descriptions too? (lines 4-64)

simulariumio/data_objects/agent_data.py
```diff
@@ -237,6 +237,7 @@ def get_display_data(type_mapping: Dict[str, Any]) -> Dict[str, DisplayData]:
             if "geometry" not in type_info:
```
schoinh, line 237: Thank you for having a doc block for every function, it's so helpful

simulariumio/data_objects/display_data.py
```diff
@@ -62,10 +62,18 @@ def __init__(
...
+            except KeyError:
+                raise DataError(
+                    f"In {name}'s DisplayData, {display_type} is not a display type, "
+                    'try "SPHERE" or see DISPLAY_TYPE documentation for other options.'
+                )
+            if not isinstance(display_type, DISPLAY_TYPE):
+                raise DataError(
+                    f"In {name}'s DisplayData, display_type is {type(display_type)}"
+                    " instead of DISPLAY_TYPE"
+                )
```
schoinh, line 76: What's an example situation where you'd NOT get a `KeyError` in line 67, proceed to line 72, and run into the error in line 73 because it's still not a valid display type?

simulariumio/tests/data/cytosim/aster_pull3D_couples_actin_solid_3_frames/aster_pull3D_couples_actin_solid_3_frames_small.json
```diff
@@ -1 +1 @@
...
...1e-6, "size": {"x": 200.0, "y": 200.0, "z": 200.0}, "typeMapping": {"1": {"name": "microtubule", "geometry": {"displayType": "FIBER"}}, "7": {"name": "motor complex"}}}, "spatialData": {"version": 1, "msgType": 1, "bundleStart": 0, "bundleSize": 3, "bundleData": [{"frameNumber": 0, "time": 0.0, "data": [1001.0, 1.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 15.0, 36.93, 36.8, 16.78, 30.55, 43.87, 19.84, 24.169999999999998, 50.93, 22.900000000000002, 17.78, 57.99999999999999, 25.95, 11.4, 65.07, 29.01, 1000.0, 12.0, 7.0, -73.8, -25.2, -43.89, 0.0, 0.0, 0.0, 2.0, 0.0]}, {"frameNumber": 1, "time": 0.05, "data": [1001.0, 1.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 18.0, 44.55, 33.97, 8.86, 36.63, 38.269999999999996, 4.51, 28.96, 43.28, 0.5, 21.83, 49.61, -2.52, 16.0, 57.66, -3.58, 12.85, 66.92, -1.47, 1000.0, 12.0, 7.0, -72.52, -21.9, -43.59, 0.0, 0.0, 0.0, 2.0, 0.0]}, {"frameNumber": 2, "time": 0.1, "data": [1001.0, 1.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 15.0, 44.55, 33.97, 8.86, 36.63, 38.269999999999996, 4.51, 28.96, 43.28, 0.5, 21.83, 49.61, -2.52, 16.0, 57.66, -3.58, 1000.0, 12.0, 7.0, -72.52, -21.9, -43.59, 0.0, 0.0, 0.0, 2.0, 0.0]}]}, "plotData": {"version": 1, "data": []}}
```
schoinh, line 1: Is this supposed to be in this PR? This is in tests/data/cytosim

simulariumio/data_objects/display_data.py
blairlyons: yeah it shouldn't be possible to get to line 72, removed that

simulariumio/tests/data/cytosim/aster_pull3D_couples_actin_solid_3_frames/aster_pull3D_couples_actin_solid_3_frames_small.json
schoinh: Gotcha. I'm still a little confused because the geometry data structure here is in the v3 format but everything else is not v3? Is it supposed to be like that? (e.g., `trajectoryInfo.version` says `1` and `spatialUnitFactorMeters` is not a v3 field)

schoinh: Ohh! Okay I was looking at an older diff :) sorry

schoinh (approved):
Looks good to me!

file_format.md
```diff
@@ -22,9 +33,13 @@ JSON files accepted by the simularium-viewer contain the following data in JSON
       * ex: "actin#barbed_ATP_1" is parsed as agent type "actin" in states "barbed", "ATP", and "1"
       * ex: "actA" is parsed as agent type "actA" with no state information
       * if no name is provided, the agent type ID, an integer number, is used for display
-    * pdb (optional) - the filename of the PDB file to render for this agent. If this field isn’t provided or if the file isn’t found, the renderer will fall back to mesh rendering
-    * mesh (optional) - the filename of the OBJ mesh file to render for this agent. If this field isn’t provided or if the file isn’t found, the renderer will fall back to a sphere
-    * PDB and mesh data is currently only used for streaming trajectories, but this will be updated soon
+    * geometry - rendering information for each agent type (note: only the first 100,000 geometry files will be loaded)
+      * displayType - “SPHERE”, “CUBE”, “GIZMO”, “FIBER”, “PDB”, or “OBJ”
```
toloudis, line 37: I believe CUBE, GIZMO, and FIBER are unsupported.  I'd remove em.  FIBER is a very confusing one, because we still have this per-agent visType which is basically fiber or not fiber.  I am really not sure if the displayable type should be per-agent or per-agenttype. If we project into the future of multiscale where fibers turn into collections of agents and back, that might shed light on what to do.  gut feeling at the moment is that the per-agent vistype should go away.

simulariumio/physicell/physicell_converter.py
```diff
@@ -38,15 +38,16 @@ def __init__(self, input_data: PhysicellData):
         self._data = self._read(input_data)
 
     @staticmethod
-    def _load_data(path_to_output_dir) -> np.ndarray:
+    def _load_data(path_to_output_dir: str, nth_timestep_to_read: int) -> np.ndarray:
```
toloudis, line 41: should there be a default value for the new arg?

file_format.md
blairlyons: removed the others, but FIBER is still a display type and a viz type afaik. Does the viewer only use the viz type? I also think it should be removed per agent and we just use the display type and subpoints, but that's a larger discussion

simulariumio/physicell/physicell_converter.py
blairlyons: this defaults to 1 in `PhysicellData` but this function isn't part of the public API and will probably only be called internally to `PhysicellConverter` which always provides it, so I didn't add it

simulariumio/tests/converters/test_physicell_converter.py
```diff
@@ -206,6 +206,162 @@
...
+                nth_timestep_to_read=2,
+                display_data={
+                    0: DisplayData(
+                        name="Cancer cell",
+                        radius=30.0,
+                        color="#0080ff",
+                    ),
+                },
+                phase_names={1: {4: "interphase"}},
+            ),
```
meganrm, line 227: I think this test is supposed to be testing the skipping frames feature, but there are so much data here, that as a developer, if this test failed, it wouldn't immediately tell me if I broke the skipping frames feature, or if some other default thing just changed and I need to update the data. It would be nice to have a separate function that tests this feature as a unit test without all the other data in the output file. Also it would make this file a lot easier to read. For instance, you don't need to keep checking cameraDefaults

blairlyons: out of scope for this PR I think, I should do that for all my tests. Can you submit a maintenance issue?

meganrm: Ok, up to you, but personally I think you could write a couple tests in a clearer format for this new feature, so that the PR includes both new feature + final tests of that new feature. Instead of placeholders that you know you're going to re-write. I'll make an issue

meganrm: https://github.com/simularium/simulariumio/issues/98

toloudis (approved):

file_format.md
toloudis: fiber is not a recognized display type in the viewer yet.   currently the viewer uses the viz type.  We have more in-depth discussion about this in other issues/PRs.

PR #95, opened 2022-03-07
Fix failing tests caused by old numpy version 

meganrm:
Problem
=======
My tests were failing once I installed cellpack. The issue was cellpack has numpy as a dependency, and it's a more recent version. But MDAnalysis is installing an old version of numpy. When the deps were being resolved in the test env, the newer version of numpy was being removed, and then the things that needed the newer version were failing the test. 

[Other people having this issue](https://stackoverflow.com/questions/66060487/valueerror-numpy-ndarray-size-changed-may-indicate-binary-incompatibility-exp)

Solution
========
I installed numpy through conda, to insure the newer version wouldn't be removed. 
I also:
1. updated some code that had numpy deprecation warnings. 
2. Use an environment.yml file to set up the conda deps, so if we want to say, install a different version of readdy, we don't have to change it in a bunch of files 

## Type of change


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -12,11 +12,11 @@ with @pairperson1
...
-- [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
-- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-- [ ] This change requires a documentation update
-- [ ] This change requires updated or new tests
+- Bug fix (non-breaking change which fixes an issue)
+- New feature (non-breaking change which adds functionality)
+- Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- This change requires a documentation update
+- This change requires updated or new tests
```
meganrm, line 19: this is just so we don't have things that look like tasks on our PRs, but it's unrelated to these other changes

simulariumio/mcell/mcell_converter.py
```diff
@@ -72,16 +72,20 @@ def _get_perpendicular_vector(v: np.ndarray, angle: float) -> np.ndarray:
...
         """
         Orthonormalize and cross the vectors to get a rotation matrix
         """
         v1 = McellConverter._normalize(v1)
         v2 = McellConverter._normalize(v2)
         v2 = McellConverter._normalize(v2 - (np.dot(v1, v2) / np.dot(v1, v1)) * v1)
         v3 = np.cross(v2, v1)
-        return np.matrix(
-            [[v2[0], v1[0], v3[0]], [v2[1], v1[1], v3[1]], [v2[2], v1[2], v3[2]]]
+        return np.array(
```
meganrm, line 83: I got a deprecation warning on np.matrix, that pointed to this information: https://numpy.org/doc/stable/user/numpy-for-matlab-users.html#array-or-matrix-which-should-i-use

simulariumio/physicell/dep/pyMCDS.py
```diff
@@ -385,15 +385,15 @@ def _read_xml(self, xml_file, parse_continuum_variables, output_path="."):
         # while we're at it, find the mesh
         coord_str = mesh_node.find("x_coordinates").text
         delimiter = mesh_node.find("x_coordinates").get("delimiter")
-        x_coords = np.array(coord_str.split(delimiter), dtype=np.float)
+        x_coords = np.array(coord_str.split(delimiter), dtype=np.float64)
```
meganrm, line 388: Deprecation warning from np: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

.github/workflows/test-and-lint.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       shell: bash -l {0}
       run: |
         conda install readdy==2.0.9
+        conda install numpy
```
toloudis: should this have the same version specifier that I see in setup.py's requirements?

schoinh (approved):

toloudis (approved):

PR #99, opened 2022-03-21
Feature/Smoldyn templates

### LINKED ISSUE #74: UI-templates for MEDYAN, Cytosim, SpringSaLaD, cellPACK
## Use Case
As a developer of the viewer, I need a template for each converter that specifies how to create a UI to gather required input data.

## Acceptance Criteria
- Each converter has a template for constructing its UI, excluding ReaDDy, PhysiCell, and MCell.

## Details
For auto-converting non-simularium files provided to the viewer, we'll need to collect some additional data from the user through a UI that will depend on the type of file they are converting. SimulariumIO is a good place to keep the specification for each converter so it can be updated with the converter.

ReaDDy, PhysiCell, and MCell don't currently accept opened file data, so they require more development work before we can support autoconversion for them.

blairlyons:
Problem
=======
The viewer needs UI templates for constructing UI to gather user input for autoconversion with each converter.

This PR only includes templates for Smoldyn, I'll add others in another PR once we verify this template setup works.
part of #74 

Solution
========
I created JSON templates in /ui-templates

## Type of change

Change summary:
---------------
* Added JSON files for templates
* Added template readme to document how the templates work

Keyfiles:
-----------------------
1. ui-templates/README.md
2. ui-templates/*.json


simulariumio/data_objects/model_meta_data.py
```diff
@@ -40,7 +40,7 @@ def __init__(
         ----------
         title : str (optional)
             Display title for this model
-        version : float (optional)
+        version : str (optional)
```
blairlyons, line 43: this was missed previously when we switched it from float to str

meganrm:
Is this supposed to still be a draft?

meganrm:
I started to try to use this on the front end and it's a bit confusing. I requested the "smoldyn_data" json, and then started to try to request the other types I need, but it's not immediately clear how to programmatically know what other files to request. For instance, if I use the keys, ie `meta_data` `display_data` etc I get some of the files correctly, but that doesn't work for `time_units` because the file name `unit_data`. But I can't use the `data_type` for requests for all of them either because there is no file named `collection`. I appreciate you wanted to keep reusable parts separate, but I think you should have a file processing step here that creates one single `smoldyn_ui_template.json` that the front end requests. Otherwise this is a lot of data wrangling to put on the front end

Short of that, it would be better to have the relative paths of the other files needed to construct the data in the json blocks

ui-templates/README.md
```diff
@@ -0,0 +1,51 @@
...
+  * parameters are indices of items in a list to be provided as a single argument to `numpy.array()`
+* `camera_data`
+* `model_meta_data`
+* `collection` - collect an extensible list of information and use it to create a dict
+  * requires additional parameter information:
+    * `length` - number of items to start with
+    * `extendible`- can the user add more items?
+    * `key_item` - requires the same fields as `parameters` items
+    * `value_item` - requires the same fields as `parameters` items
+* `display_data`
```
meganrm, line 45: you call `display_data` a data type here, but in the `smoldyn_data`, display_data's `data_type` is set to `collection`

blairlyons:
I removed the "default" field and moved that info to "help". This means the help should now be displayed to the user before they enter anything, regardless of whether the value fails validation.

@meganrm TODO from our meeting: move reuseable objects to a sub directory (possibly with metadata about what types are available) that the FE can download regardless of engine type

ui-templates/README.md
blairlyons: `display_data` is the name of a parameter on the `SmoldynData` object that is a collection of `DisplayData`s for each agent type. It's confusing that it has the same name as a template type, but I don't think it makes sense to change either of them

toloudis: It's reasonable to disallow the reuse of data type names as data value names. But at the same time, we usually know where in the data we expect to see a type name vs a value name. In programming languages, usually you have a convention to name `DataType` with different casing than a `data_value`(Python) or `dataValue`(TS/JS/C++) of that type.  

Can you have `collection`s of different data types or is a collection always expected to be of the same type?  If always of a consistent type, you could replace `collection` with `collection[display_data]` or `display_data[]` shorthand notation that is easy enough to parse to see that it's intended to be a collection.  Just a thought but don't know if it's viable with this code.

ui-templates/base_types.json
```diff
@@ -0,0 +1,30 @@
...
+            "data": "string",
+            "match": "/#(([0-9a-fA-F]{2}){3}|([0-9a-fA-F]){3})/i"
+        },
+        {
+            "id": "collection",
+            "data": "array"
+        },
+        {
+            "id": "enum",
+            "data": "string"
```
meganrm, line 27: should be `string[]`

ui-templates/custom-types/unit_data.json
```diff
@@ -0,0 +1,21 @@
+{
+    "unit_data": {
+        "python::module": "simulariumio",
+        "python::object": "UnitData",
+        "parameters": {
+            "name": {
+                "name": "Name",
+                "data_type": "text",
+                "description": "Provide the unit name (we support this list https://github.com/hgrecco/pint/blob/master/pint/default_en.txt)",
```
meganrm, line 9: shouldn't this be data_type `enum` instead of `text`?

ui-templates/custom-types/unit_data.json
```diff
@@ -0,0 +1,21 @@
...
+                "required": true
+            },
+            "magnitude": {
+                "name": "Magnitude",
+                "data_type": "number",
+                "description": "Provide a multiplier in case values are not in whole units",
+                "required": false,
```
meganrm, line 16: so it's ok to make `UnitData({name: "nm"})`? I feel like i've seen it `UnitData("nm")`

ui-templates/base_types.json
meganrm: (this is a note to myself)

meganrm:
@blairlyons I managed to get the paths to work, you don't need to change the coordinate x y z unless you want to

ui-templates/smoldyn_data.json
```diff
@@ -0,0 +1,50 @@
...
+        "python::object": "SmoldynData",
+        "parameters": {
+            "meta_data": {
+                "name": "Metadata",
+                "data_type": "meta_data",
+                "description": "Provide metadata for this trajectory",
+                "required": false,
+                "help": "If not provided, defaults to box size = [100, 100, 100], scale = 1.0, and no metadata about the trajectory or model"
+            },
+            "display_data": {
```
meganrm, line 13: I think we talked about this before, but this is still confusing to me that there is both a parameter called "display_data" and a "data_type": "display_data". Can we get different names for these?

ui-templates/custom-types/coordinates_xyz.json
```diff
@@ -0,0 +1,26 @@
+{
+    "coordinates_xyz": {
+        "python::module": "numpy",
+        "python::object": "array",
+        "parameters": {
+            "0": {
+                "name": "X",
+                "data_type": "number",
+                "description": "Provide the X dimension",
+                "required": true
```
meganrm, line 10: The use of required here is a little odd, because it's parent is not required. I was figuring I would use this required field to add the little red *, but it doesn't really make sense here because it makes it seem like the whole thing is required. Maybe we can figure out a different way to denote that if you enter one of the subfields you need them all. 
Screen Shot 2022-06-29 at 10 37 07 AM

ui-templates/custom-types/unit_data.json
meganrm: I *think* it errored out when I tried the first thing, but something else may have gone wrong. do you have a test that does it both ways?

ui-templates/custom-types/unit_data.json
meganrm: Like you did here: 
```     
 "options": [
                    "SPHERE",
                    "PDB",
                    "OBJ",
                    "FIBER"
                ]
```

ui-templates/README.md
```diff
@@ -0,0 +1,50 @@
...
+  * `help` - additional information to help the user, including what default value will be used (optional)
+  * some types also have additional parameter information
+
+### Main types
+* `smoldyn_data` - information for populating a `SmoldynData` to visualize a trajectory from Smoldyn
+* (more coming soon)
+
+### Base types
+
+* `number`
```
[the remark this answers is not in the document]
meganrm: resolution: enforce that the values are always floats. Use the word "float" here instead @blairlyons

ui-templates/custom-types/unit_data.json
meganrm: @blairlyons  change this file to two files, one for time and one for space, which each have options of the available SI units for time and space

ui-templates/smoldyn_data.json
```diff
@@ -0,0 +1,50 @@
+{
+    "smoldyn_data": {
+        "python::module": "simulariumio.smoldyn",
+        "python::object": "SmoldynData",
+        "parameters": {
+            "meta_data": {
```
meganrm, line 6: @blairlyons add a `file_contents` here that is a base type="file"

ui-templates/README.md
```diff
@@ -0,0 +1,51 @@
...
+    * `length` - number of items to start with
+    * `extendible`- can the user add more items?
+    * `key_item` - requires the same fields as `parameters` items
+    * `value_item` - requires the same fields as `parameters` items
+* `display_data`
+* `enum`
+  * requires additional parameter information:
+    * `python::module` - Python module for the enum
+    * `python::object` - Python name of the enum
+    * `options` - enum values that a user can choose
```
blairlyons, line 50: the display_type options actually map to an enum in Python, so the `enum` type requires `python::module` and `python::object`. The new unit_data options do not map to an enum, they are just strings. Will that be ok? seems potentially confusing.

ui-templates/README.md
```diff
@@ -0,0 +1,52 @@
...
+### Base types
+
+* `float`
+* `text`
+* `url`
+* `color`
+
+### Custom types
+
+* `meta_data`
```
toloudis, line 34: do you intend to document these other fields here?  some have notes and some don't

toloudis: for example, what's the difference between meta_data and model_meta_data?

ui-templates/smoldyn_data.json
```diff
@@ -0,0 +1,56 @@
...
+                "name": "Display data",
+                "data_type": "collection",
+                "description": "Provide display information for each particle type from Smoldyn",
+                "required": false,
+                "help": "If not provided, defaults to display name from Smoldyn, and draws agent as a sphere with radius = 1.0 and default colors",
+                "length": 1,
+                "extendible": true,
+                "key_item": {
+                    "name": "Particle name",
+                    "data_type": "text",
```
toloudis, line 29: are there collections that ever have keys that are not text?
Also, isn't this display data a very standard form that will show up everywhere?  this whole collection thing could become a custom-type, right?

toloudis (approved):
I think I've already said all my concerns with this design but I'll write em here.

I'm slightly worried that this code is trying to do too much with json just for the sake of making things completely data driven and having to "change things only in one place".

On the UI side I still think we would get more flexibility for layout and rendering if we can let things like spatial units or xyz vectors be their own React components instead of forcing them to be built entirely data-driven.

It's not quite clear from the PR how this will connect up with python.
Are all of the `python::module` and `python::object` already written in the Python code?
Do you require a 1-1 correspondence between the Python object fields and the json fields?
We talked about having a functional relationship where the json would then not need to have the exact names as the python properties or it wouldn't have to be 1-1 if that need ever arose. It would also serve to keep the code better encapsulated. (e.g. python class has a constructor or init function that can take in a whole json object and then be responsible for extracting the fields on its own)

blairlyons:
> I think I've already said all my concerns with this design but I'll write em here.
> 
> I'm slightly worried that this code is trying to do too much with json just for the sake of making things completely data driven and having to "change things only in one place". I'm not convinced that anything is actually wrong what's written here, but what is implied when this code develops further and is glued together with other parts.
> 
> On the UI side I still think we would get more flexibility for layout and rendering if we can let things like spatial units or xyz vectors be their own React components instead of forcing them to be built entirely data-driven.
> 
> It's not quite clear from the PR how this will connect up with python. Are all of the `python::module` and `python::object` already written in the Python code? Do you require a 1-1 correspondence between the Python object fields and the json fields? We talked about having a functional relationship where the json would then not need to have the exact names as the python properties or it wouldn't have to be 1-1 if that need ever arose. It would also serve to keep the code better encapsulated. (e.g. python class has a constructor or init function that can take in a whole json object and then be responsible for extracting the fields on its own)

@toloudis @meganrm 
I think we should talk about this more before calling this done. Maybe we should just go with the current design and iterate, we will have more info once we implement the full data-driven conversion (https://github.com/simularium/simularium-lambdas/issues/1) and add support for more engines beyond Smoldyn, but maybe we know enough to rethink some of it already?

meganrm (approved):

ui-templates/custom-types/display_data.json
```diff
@@ -0,0 +1,53 @@
...
+            },
+            "display_type": {
+                "name": "Display type",
+                "data_type": "enum",
+                "description": "type of geometry",
+                "required": true,
+                "python::module": "simulariumio",
+                "python::object": "DISPLAY_TYPE",
+                "options": [
+                    "NONE",
```
meganrm, line 27: @blairlyons remove this

Issue #98, opened 2022-03-21
Rewrite converter tests 


meganrm:
## What needs to happen?
Write clearer unit tests that test the particular functionality of the converter, with smaller chunks of expected data. 
Make sure both the default settings, and all the optional input data are covered in the tests. 

The cellPACK converter tests are a good example of how we want to break up tests https://github.com/simularium/simulariumio/blob/main/simulariumio/tests/converters/test_cellpack_converter.py

## Why should we do this?
Currently, the tests cover so much of the data, that if one thing changed, all the tests would fail, instead of just the tests that would tell you what particular thing had changed, so has a developer, it would be challenging to figure out if you had introduced a bug or if you just need to update the expected data in the test. It would also make the test files a lot easier to read.

PR #105, opened 2022-06-28
Feature/metaballs

### LINKED ISSUE #96: Metaballs display type
## Use Case
As a user with agents that change shape over time, I want to prepare my data to visualize these agents without having to prepare meshes for each agent at each time step.

## Acceptance Criteria
Simularium files have a metaballs display type, where positions and radii of metaballs are saved in subpoints for that agent.

## Details
- add a display type "METABALLS"
- save XYZ position and radius for each metaball in the subpoints for the agent, i.e. [metaball1 position X, metaball1 position Y, metaball1 position Z, metaball1 radius, metaball2 position X, metaball2 position Y, metaball2 position Z, metaball2 radius, ...]
- also add a viz type "METABALLS"? Or is now a good time to talk about removing viz type since it is redundant with display type?

blairlyons:
Problem
=======
Agents that change shape over time had to be saved as meshes per agent per timestep, now they can optionally be rendered as groups of spheres instead
closes #96 

Solution
========
Add a `SPHERE_GROUP` display type and save sphere positions and radii in `subpoints`.

coordinated with @toloudis for matching implementation in https://github.com/simularium/simularium-viewer/issues/255

## Type of change
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

**Since binary PR is still open, this diff is more useful https://github.com/simularium/simulariumio/compare/feature/binary..feature/metaballs

Change summary:
---------------
* add a `SPHERE_GROUP` display type and save sphere positions and radii in `subpoints`.
* change `AgentData.subpoints` to be flat per agent, it had been assuming a shape of (n, 3) since subpoints were only used for fiber points
* make `display_type` required for a `DisplayData` since it's no longer easy to choose a default based on subpoints
* add a `display_data` argument to create a `FileConverter` so old file versions can be read, since display data is now required
* add `SPHERE_GROUP` output options to `PhysicellConverter` and `PhysicellData`, including coloring by agent ID
* update filters that deal with subpoints
* update tests and example notebooks

Keyfiles:
-----------------------
1. file_format.md - add `SPHERE_GROUP`
2. simulariumio/constants.py - add `VIZ_TYPE.SPHERE_GROUP`


toloudis:
should this be targeted for the feature/binary branch instead of main?  105 files diffed in this PR??

blairlyons:
The binary PR should merge first, and then both go to main, but yes I just changed this to target feature/binary so the diff is useful @toloudis

LICENSE
```diff
@@ -1,36 +1,201 @@
...
-WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
-DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
-FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
-DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
-SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
-OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+                                 Apache License
+                           Version 2.0, January 2004
```
toloudis, line 2: out of scope of metaballs PR?

simulariumio/constants.py
```diff
@@ -24,7 +24,7 @@ class V1_SPATIAL_BUFFER_STRUCT:
     R_INDEX: int = 9  # radius
     NSP_INDEX: int = 10  # num subpoints
     SP_INDEX: int = 11  # subpoints
-    VALUES_PER_AGENT: int = 12
+    VALUES_PER_AGENT: int = 11
```
toloudis, line 27: why is this changing?

simulariumio/springsalad/springsalad_converter.py
```diff
@@ -96,11 +102,15 @@ def _parse_springsalad_data(
                 result.n_agents[time_index] += 1
                 result.unique_ids[time_index][agent_index] = int(cols[1])
                 raw_type_name = cols[3]
-                result.types[time_index].append(
-                    input_data.display_data[raw_type_name].name
-                    if raw_type_name in input_data.display_data
-                    else raw_type_name
-                )
+                if raw_type_name not in input_data.display_data:
```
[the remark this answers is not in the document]
toloudis: display_type_name = get_display_type_name_from_raw(raw_type_name, input_data.display_data)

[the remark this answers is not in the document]
toloudis: and put all the if-else logic inside that func

simulariumio/constants.py
toloudis: This is the only scary one to me.  It really looks like the number of values per agent should be 12.

blairlyons: 12 included the first subpoint value. If the number of subpoints is 0, there will be no subpoint values. In a lot of the code I updated, it had been implemented using `VALUES_PER_AGENT - 1`

simulariumio/constants.py
```diff
@@ -63,10 +63,23 @@ class DISPLAY_TYPE(Enum):
...
+    """
+    How many subpoints per item saved in subpoints?
+    given the display type of the agent
+    """
+    if display_type == DISPLAY_TYPE.FIBER:
+        return 3
+    if display_type == DISPLAY_TYPE.SPHERE_GROUP:
+        return 4
+    return 1
+
```
blairlyons, line 81: this is necessary now that subpoints can either contain fiber points or sphere group points and radii

LICENSE
blairlyons: had to fix the diff since binary has been merged

simulariumio/filters/translate_filter.py
```diff
@@ -67,24 +66,29 @@ def apply(self, data: TrajectoryData) -> TrajectoryData:
...
+                    max_subpoints > 0 and display_type != DISPLAY_TYPE.SPHERE_GROUP
+                )
+                if translate_subpoints:
+                    sp_items = self.get_items_from_subpoints(
+                        data.agent_data, time_index, agent_index
+                    )
+                    if sp_items is None:
+                        translate_subpoints = False
+                    else:
+                        # translate subpoints for fibers
```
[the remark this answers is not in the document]
blairlyons: I think we ideally want fiber points to be relative to the fiber agent's position, but the viewer wasn't rendering it that way, at least partially because of a bug. I made an issue for it, I think the viewer can handle that now? https://github.com/simularium/simulariumio/issues/109

simulariumio/springsalad/springsalad_converter.py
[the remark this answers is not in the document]
blairlyons: I looked into doing this, and the issue with having a shared function is that for most of the converters, they have a specific way of setting up the `DisplayData` (including the `display_type_name`) given the `raw_type_name`, which is what's happening in the `if` clause here. I thought about trying to make an abstract method or passing a handler for that part and letting each converter implement it differently, but they mostly use different pieces of additional data to do it, so would need different arguments. It starts to seem like a lot of work to save a redundant `if` loop. But I can separate this into a function in each of the converters, it's a lot of logic to read directly in this main loop.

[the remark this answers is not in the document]
blairlyons: @toloudis I made a function on `TrajectoryConverter` that a few of them can share (MCell, Smoldyn, SpringSaLaD), otherwise I just separated them into a new (unshared) function for readability, if it made sense.

simulariumio/tests/filters/test_translate_filter_sphere_groups.py
```diff
@@ -0,0 +1,1213 @@
...
+                                -0.7327957,
+                                0,
+                                0,
+                                0,
+                                1.0,
+                                4,
+                                0,
+                                0,
+                                0,
+                                0.1882706,
```
meganrm, line 1199: does this data really need to be this large to demonstrate the test is working? It makes it impossible to intuitively see what this test is doing.

meganrm:
All the stuff in phyiscell/output seems like it should be git ignored?

simulariumio/tests/conftest.py
```diff
@@ -131,9 +131,11 @@ def three_default_agents() -> TrajectoryData:
             display_data={
                 "C": DisplayData(
                     name="C",
+                    display_type=DISPLAY_TYPE.SPHERE,
```
meganrm, line 134: Is this change because you removed the default setting?

meganrm:
I think I need a walkthrough of this, I don't understand this statement: "change AgentData.subpoints to be flat per agent, it had been assuming a shape of (n, 3) since subpoints were only used for fiber points"

simulariumio/medyan/medyan_converter.py
```diff
@@ -103,13 +108,13 @@ def _parse_data_dimensions(
...
                     cols = line.split()
-                    subpoints = int(cols[3])
+                    subpoints = 3 * int(cols[3])
                     if result.max_subpoints < subpoints:
                         result.max_subpoints = subpoints
                 else:
                     # start of linker or motor
-                    if result.max_subpoints < 2:
-                        result.max_subpoints = 2
+                    if result.max_subpoints < 6:
```
meganrm, line 116: I currently don't understand what's happening here, but even once I do understand, this seems like an if statement that could use an explanatory comment

simulariumio/medyan/medyan_converter.py
```diff
@@ -189,9 +196,9 @@ def _get_trajectory_data(input_data: MedyanData) -> AgentData:
                     input_data.meta_data.scale_factor * radius
                 )
                 if object_type == "filament":
-                    result.n_subpoints[time_index][agent_index] = int(cols[3])
+                    result.n_subpoints[time_index][agent_index] = 3 * int(cols[3])
                 else:
-                    result.n_subpoints[time_index][agent_index] = 2
+                    result.n_subpoints[time_index][agent_index] = 6
```
meganrm, line 201: it seems like `6` could maybe be a named constant. that might help with clarity

simulariumio/constants.py
```diff
@@ -24,7 +24,7 @@ class V1_SPATIAL_BUFFER_STRUCT:
     R_INDEX: int = 9  # radius
     NSP_INDEX: int = 10  # num subpoints
     SP_INDEX: int = 11  # subpoints
```
toloudis, line 26: change "subpoints" to "index of first subpoint, only valid if num_subpoints > 0"

simulariumio/tests/conftest.py
blairlyons: yes

simulariumio/constants.py
```diff
@@ -24,7 +24,7 @@ class V1_SPATIAL_BUFFER_STRUCT:
...
-    VALUES_PER_AGENT: int = 12
```
blairlyons, line 27: name this `MIN_PER_AGENT` or something

simulariumio/constants.py
blairlyons: add comment about what 3 and 4 mean, or create symbolic constants

simulariumio/data_objects/display_data.py
```diff
@@ -23,8 +23,8 @@ class DisplayData:
     def __init__(
         self,
         name: str,
+        display_type: DISPLAY_TYPE,
```
blairlyons, line 26: add error check to make sure number of subpoints corresponds to display_type

simulariumio/medyan/medyan_converter.py
blairlyons: I added symbolic constants and more comments

simulariumio/tests/filters/test_translate_filter_sphere_groups.py
blairlyons: I shortened this new test to only check the bundle data for 1 frame, since this filter only changes the positions and subpoints. In order to run filter() I need a converter, so still have to create a full trajectory as input to the test, but the expected data is now only the bundle data for 1 frame. 

Breaking up the other tests that I updated in this PR will be done in https://github.com/simularium/simulariumio/issues/98

simulariumio/cytosim/cytosim_converter.py
```diff
@@ -71,7 +77,7 @@ def _parse_object_dimensions(
                     agents += 1
                 continue
             if is_fiber:
-                subpoints += 1
+                subpoints += VALUES_PER_3D_POINT
```
toloudis, line 80: I'm not sure but it looks like this could be using SUBPOINT_VALUES_PER_ITEM ?

toloudis (approved):

simulariumio/cytosim/cytosim_converter.py
blairlyons: yep, fixed it a few other places too

meganrm (approved):

PR #110, opened 2022-07-23
Update PhysiCell importer

blairlyons:
Problem
=======
PhysiCell has made updates to their import script and we want to pull them in to stay compatible with new PhysiCell output files.

Solution
========
Since PhysiCell has a Python script for import but not a Python package, we have our own copy of their script in `simulariumio/physicell/dep/`. I just copied in their new updates.

We want to merge this ahead of our session at their workshop on Tuesday next week.

## Type of change

Change summary:
---------------
* update pyMCSD.py

### DIFF
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -62,5 +62,5 @@ jobs:
     - name: Lint with flake8
       shell: bash -l {0}
       run: |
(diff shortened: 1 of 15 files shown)


toloudis (approved):
I don't know the file format so it's hard for me to judge correctness but I don't see anything obviously scary.

meganrm (approved):

PR #113, opened 2022-08-08
Feature/particle orientations

### LINKED ISSUE #40: Calculate particle topologies orientations
## Use Case
It would be nice if users with topologies of connected particles could visualize them with geometry, but they need to calculate orientations for the particles first.

## Acceptance Criteria
After providing orientation parameters, a user with connected particles can calculate orientations as part of the visualization process.

## Details
In simularium-models-util `ActinUtil.get_actin_rotation()` already has an implementation of this (https://github.com/allen-cell-animated/simularium-models-util/blob/7aca5dfa3c9617c4f19666c94428047f3ed8818f/simularium_models_util/actin/actin_util.py#L92) but it needs to be generalized to be used with any molecule combinations.

blairlyons:
Problem
=======
Particles can only be drawn with spheres because the particles have no orientations. 
closes #40 

Solution
========
Particles with geometrically-constrained edges (e.g. in ReaDDy data) can be mapped to structural data (e.g. from PDB files) to back-calculate orientations that allow geometry generated from that structural data to be drawn with reasonable accuracy.

This solution breaks the problem into multiple cases:
1. Independent orientations - particles with two neighbors have fully determined orientations. To calculate them, use the vectors from the particle to each of its neighbors to form a [basis](https://en.wikipedia.org/wiki/Basis_(linear_algebra)).
2. Random orientation - if a particle has no neighbors, it's orientation is fully undetermined, and the only thing to do is create a random orientation (**Note**: maybe a future PR should make the random rotation be an offset from the previous time step, for now it is independently random each time. However this might be roughly accurate given these timesteps are very large relative to molecular time).
3. Dependent orientation - particles with one neighbor may have fully or partially determined orientations:
  - If their one neighbor has other neighbors, their orientation is fully determined by an offset from their neighbor's orientation
  - If their one neighbor has no other neighbors, both of these particles have to orient correctly relative to each other, but have a degree of freedom around the axis defined by their edge. (This case also applies to colinear neighbors but I've left that out for now)

For all of these cases, the orientations calculated from neighbor positions are meaningless on their own. In order to use them to orient geometry, we need to calculate the difference between the current orientation and the orientation of the particles used to generate the geometry (which I'm calling "zero orientations").

## Type of change

Change summary:
---------------
* implement orientations for two neighbors (independent)
* implement orientations for one neighbor (dependent)
* use orientations in ReaDDy converter
* verify rotations look reasonable in the viewer
* add tests

Keyfiles:
-----------------------
1. simulariumio/orientations/particle_rotation_calculator.py - `ParticleRotationCalculator` is the main particle orientation class, which contains all the data needed to calculate orientation for one particle at one time step. In `__init__()` it calculates independent and random orientations since they don't depend on neighbor orientations already being calculated. It also has `calculate_dependent_rotation()` which can be used after `ParticleRotationCalculator`s are created for all the particle's neighbors. There's also `get_euler_angles()` to get the final orientation as an XYZ rotation for the simularium viewer.
2. simulariumio/readdy/readdy_converter.py and simulariumio/readdy/readdy_data.py - add orientation calculation for ReaDDy particles.
3. simulariumio/orientations/orientation_data.py and simulariumio/orientations/neighbor_data.py - data classes with some related methods.
4. simulariumio/orientations/rotation_utility.py - helpers for calculating rotations, everything that uses scipy for linear algebra and rotations


simulariumio/orientations/rotation_utility.py
```diff
@@ -0,0 +1,209 @@
...
+        return np.array(
+            [
+                [vector1[0], vector2[0], vector3[0]],
+                [vector1[1], vector2[1], vector3[1]],
+                [vector1[2], vector2[2], vector3[2]],
+            ]
+        )
+
+    @staticmethod
+    def get_random_rotation_matrix():
```
blairlyons, line 131: use scipy rotation?

simulariumio/orientations/__init__.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from .particle_rotation_calculator import ParticleRotationCalculator  # noqa: F401
+from .orientation_data import OrientationData  # noqa: F401
+from .neighbor_data import NeighborData  # noqa: F401
+from .rotation_utility import RotationUtility  # noqa: F401
```
toloudis, line 7: IMO, everything in this directory should really be a subdirectory of `readdy` (except maybe rotation_utility).  I'd just move the other 3 files into `readdy/orientations`.  All the data structures are so custom to readdy for now.

Also wondering if there is code to generate OrientationData from pdb file.  That code would be great to have in this repo really well.

simulariumio/orientations/rotation_utility.py
toloudis: or generate from random euler angles using your get_rotation_matrix_from_euler_angles... anything to simplify this to a one-liner that's super readable.

simulariumio/orientations/rotation_utility.py
```diff
@@ -0,0 +1,209 @@
...
+import random
+
+###############################################################################
+
+log = logging.getLogger(__name__)
+
+###############################################################################
+
+
+class RotationUtility:
```
toloudis, line 20: does this really need to be a class?  It could just be a plain old Python module and the function calls namespaced to the import itself.

simulariumio/orientations/__init__.py
blairlyons: there's a procedure I use in chimera, I can add it to the documentation at least. It's all things that could be a script but aren't yet.

simulariumio/orientations/rotation_utility.py
blairlyons: It makes imports nicer? so I don't have to export all the functions that are available in __init__.py. Is there a nice way to do that without a class?

simulariumio/orientations/particle_rotation_calculator.py
```diff
@@ -0,0 +1,299 @@
...
+        self.current_rot_matrix = RotationUtility.get_rotation_matrix_from_bases(v1, v2)
+
+    def _match_orientation_data_with_one_neighbor(
+        self,
+        particle_type_name: str,
+        neighbor_type_names: List[str],
+    ) -> Tuple[OrientationData, int, int]:
+        """
+        Set the matching orientation data
+        for the particle and neighbor of the given type
```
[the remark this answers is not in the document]
blairlyons: I added a comment for this, it's definitely not the most intuitive. Basically it's returning the indices of the matching neighbors, if there was only one neighbor, it returns -1 for the neighbor position that was not matched so later code can tell which piece neighbor's data to use. If you have ideas to make this better I'm very open to that!

ascibisz (approved):
I don't understand all of the math, but from what I do understand it looks good to me!

simulariumio/orientations/rotation_utility.py
toloudis: I think it's something like `from rotation_utility import *` or just `import rotation_utility` to maintain the module namespacing like I wrote above.

simulariumio/orientations/particle_rotation_calculator.py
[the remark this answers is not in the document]
toloudis: commenting it is a good start.  
namedtuples, or data classes is the next step up.   basically a structure where you can give semantic names to the tuple items for readability.  A slightly worse alternative is to do a typedef where you alias `int` to `NeighborIndex`

toloudis (approved):

simulariumio/orientations/rotation_utility.py
blairlyons: I know I could do `from rotation_utility import *` but then it can be hard to tell where the functions came from, I tried just `import rotation_utility` and maybe I'm doing that wrong but can't get it to work... so I think I'm going to leave it as a class.

PR #114, opened 2022-08-22
admin/update-to-new-cookiecutter

evamaxfield:
Problem
=======
What is the problem this work solves, including
[Link to story or ticket](https://my-tracking-system.url/ticket-number)

@blairlyons asked me to try and update to the new cookiecutter I made: https://github.com/evamaxfield/cookiecutter-py-package

Solution
========
What I/we did to solve this problem

I ran the upgrade and debugged everything I could.

Will debug CI as it fails (but I hope it works out-of-the-gate).

Note: docs uses a new theme.
image

image

image

----

Some comments:

The cookiecutter I maintain is meant to build packages to py310 and it has formatting with `black` and `isort` automatically plus type-checking.

Those four things were the largest culprits of failure.
* Py310 isn't supported in this library because of upstream `cellpack` dependency which depends on `panda3d`.
* `black` and `isort` are good for formatting. @toloudis can back me up here :woman_shrugging:  but if I had run them in this PR, the PR would have changed ~80 files
* `mypy` / type checking is cool. But a lot to debug without the context of the library. Add it whenever you want if you want the extra "statically typed check" safety measure (I would argue that anything IO based should add type checking because FileIO errors are hard for users to understand :woman_shrugging:)

There were some other specific changes _from the cookiecutter_ I made to make this work.
The `just install` command creates a conda env and then installs readdy and numpy from conda-forge along with installing all the deps.

If you look at the `Justfile` for `just install` you will see it does a VERY long install string. Technically this is "safer" than the `pip install -e .[dev]` we used to do because we have more _explicit_ control over the installs but it is annoying, so we just have a just command for it.

`just test` replaces `tox`.
`just lint` / `pre-commit` replaces `tox`
`just build` replaces `tox`

Let me know if you have questions


.cookiecutter.yaml
```diff
@@ -0,0 +1,10 @@
+default_context:
+    author_name: "Blair Lyons"
+    author_email: "blairl@alleninstitute.org"
+    hosting_github_username_or_org: "Simularium"
+    project_name: "SimulariumIO"
+    project_slug: "simulariumio"
+    python_slug: "simulariumio"
+    project_short_description: "Python package that converts simulation outputs to the format consumed by the Simularium viewer website"
+    hosting_pypi_username: "__token__"
+    license: "Apache Software License 2.0"
```
evamaxfield, line 10: note to self: I need to add the Apache license to the cookiecutter options.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
```
evamaxfield, line 1: Makes sure that your github actions are up-to-date (usually good for security)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,111 @@
+name: CI
```
evamaxfield, line 1: No more separate workflows for build-main and test-and-lint

This does both! And handles publishing

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+files: simulariumio
```
evamaxfield, line 1: If you wanted to add `black` `flake8` and `mypy`, this is where they would go. Let me know when you want to do that.

CONTRIBUTING.md
```diff
@@ -31,45 +38,56 @@ Ready to contribute? Here's how to set up `simulariumio` for local development.
     Now you can make your changes locally.
 
 5. When you're done making changes, check that your changes pass linting and
-   tests, including testing other Python versions with make:
+   tests with [just](https://github.com/casey/just):
```
evamaxfield, line 41: `just` is _just_ a better `make`. Fancy :sparkles: written in Rust. But the real advantages imo are two:
1. variable handling is WAY easier and it has custom functions we can call for certain things like `{{justfile_directory()}}` to get the directory of the `Justfile` itself or `{{join()}}` for path joining on all OSes.
2. It runs from any subdir! `make` must be run in the dir where the `Makefile` is, but `just` will search up directories until it finds one.

CONTRIBUTING.md
```diff
@@ -31,45 +38,56 @@ Ready to contribute? Here's how to set up `simulariumio` for local development.
...
-## Questions or Thoughts?
+A reminder for the maintainers on how to deploy.
+Make sure the main branch is checked out and all desired changes
+are merged. Then run:
 
-Talk to us on [one of our community forums][community].
+```bash
+just tag-for-release "vX.Y.Z"
+just release
+```
```
evamaxfield, line 88: Versions are entirely controlled by git tags! It means that even when you pip install editable mode (`just install`) it will have like a `3.1.2.dev{git_hash}`!

Justfile
```diff
@@ -0,0 +1,76 @@
...
+project_uri := if "os_family()" == "unix" {
+	justfile_directory()
+} else {
+	replace(justfile_directory(), "\\", "/")
+}
+
+# generate Sphinx HTML documentation and serve to browser
+serve-docs:
+	just generate-docs
+	python -mwebbrowser -t "file://{{project_uri}}/docs/_build/index.html"
```
evamaxfield, line 62: This is why Justfile is better imo.

benchmarks/benchmark_simulariumio.py
```diff
@@ -6,18 +6,11 @@
...
-    CytosimObjectInfo,
-)
-from simulariumio.springsalad import (
-    SpringsaladConverter,
-    SpringsaladData,
-)
-from download_benchmark_resources import download_benchmark_resources, Args
+from simulariumio import DISPLAY_TYPE, DisplayData, InputFileData, MetaData
+from simulariumio.cytosim import CytosimConverter, CytosimData, CytosimObjectInfo
+from simulariumio.springsalad import SpringsaladConverter, SpringsaladData
```
evamaxfield, line 13: Accidental `isort` before I removed it.

docs/index.rst
```diff
@@ -1,4 +1,4 @@
-Welcome to Simularium Conversion's documentation!
+Welcome to simulariumio's documentation!
```
evamaxfield, line 1: I removed a bunch of these docs/*rst files because they are auto-generated and we don't need to track them in git.

pyproject.toml
```diff
@@ -0,0 +1,171 @@
+# package build
```
evamaxfield, line 1: The new package manager doc.

Most things have a notes as to what they do!

pyproject.toml
```diff
@@ -0,0 +1,171 @@
...
+]
+classifiers = [
+  "Development Status :: 4 - Beta",
+  "Natural Language :: English",
+  "Programming Language :: Python :: 3.7",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+]
+dynamic = ["version"]
+dependencies = [
```
evamaxfield, line 29: main deps

pyproject.toml
```diff
@@ -0,0 +1,171 @@
...
+
+[project.urls]
+Homepage = "https://github.com/Simularium/simulariumio"
+"Bug Tracker" = "https://github.com/Simularium/simulariumio/issues"
+Documentation = "https://Simularium.github.io/simulariumio"
+"User Support" = "https://github.com/Simularium/simulariumio/issues"
+
+# extra dependencies
+# https://peps.python.org/pep-0621/#dependencies-optional-dependencies
+[project.optional-dependencies]
```
evamaxfield, line 43: optional deps

pyproject.toml
```diff
@@ -0,0 +1,171 @@
...
+# build settings
+# https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html
+[tool.setuptools]
+zip-safe = false
+include-package-data = true
+
+[tool.setuptools.packages.find]
+exclude = ["*docs/*", "*tests/*"]
+
+[tool.setuptools.package-data]
```
evamaxfield, line 112: if you want to add data to ship with the package to `pypi`

blairlyons:
I tried just editing ci.yml to remove Python 3.7 (we don't support it anymore, checks failed), but now checks aren't running. Do we need to add something to make the workflow run when PR is edited not just opened? @evamaxfield @ascibisz

evamaxfield:
> I tried just editing ci.yml to remove Python 3.7 (we don't support it anymore, checks failed), but now checks aren't running. Do we need to add something to make the workflow run when PR is edited not just opened? @evamaxfield @ascibisz

Iirc its because of two things:

1. There are now merge conflicts
2. Because I am a new contributor to this repo, after the merge conflicts are fixed, you have to manually click like "Run Actions" or something. It will pop up at the bottom of the PR page after merge conflicts are gone.

A bit busy today. I can fix merge conflicts tomorrow or you can push to my branch

docs/index.rst
meganrm: Thank you! Those have bothered me

.github/workflows/ci.yml
```diff
@@ -0,0 +1,121 @@
...
+    - name: Install Dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install .[lint]
+    - name: Lint
+      run: just lint
+
+  # Publish to PyPI if test, lint, and manifest checks passed
+  publish:
+    if: "success() && startsWith(github.ref, 'refs/tags/')"
```
meganrm, line 98: why do you use the "" around this? I don't think you need to

evamaxfield: I think this is just a mistake in the cookiecutter template. I have no idea why it's there because it seems like the github docs would agree that no string wrapper is necessary :woman_shrugging: https://docs.github.com/en/actions/learn-github-actions/expressions#example-10

I will remove from cookiecutter and pull in that change.

meganrm (approved):

ascibisz (approved):

Issue #78, opened 2021-11-17
Check unique IDs are not too large


blairlyons:
## Use Case
Agent unique IDs greater than the integer limit (~2e9) cause the viewer to crash, so it would be nice if SimulariumIO validated them first.

## Acceptance Criteria
If any agent unique IDs are greater than the integer limit (~2e9), SimulariumIO throws an error during conversion.

## Details
Would also be good to check that any code generating agent unique IDs is doing so sensibly, e.g. start the IDs over at zero every timestep.

PR #116, opened 2022-08-25
Bugfix/validate agent id size

ascibisz:
Problem
=======
Simulariumio should error if Agent IDs are too large
[#78](https://github.com/simularium/simulariumio/issues/78)

Solution
========
If agent IDs are larger than a 32-bit signed int, throw a data error instead of writing to the .simulariumio file. This check is optional, and users can save with the `validate_ids` argument set to `False` if they want to avoid the additional validation step in order to improve performance.

I also manually verified that all of our converters are generating agent IDs sensibly and not continuing to increment agent IDs unnecessarily in subsequent time steps.

## Type of change

Change summary:
---------------
* Add check in writer to verify that all agent IDs are small enough to be represented as a 32 bit signed int
* Add test to ensure check is working properly

### DIFF
--- a/simulariumio/constants.py
+++ b/simulariumio/constants.py
@@ -127,3 +127,6 @@ def JMOL_COLORS() -> pd.DataFrame:
     """
     this_dir, _ = os.path.split(__file__)
     return pd.read_csv(os.path.join(this_dir, JMOL_COLORS_CSV_PATH))
+
+
+MAX_AGENT_ID = 0x80000000  # Agent IDs should fit in a 32 bit signed int
--- a/simulariumio/tests/converters/test_trajectory_converter.py
+++ b/simulariumio/tests/converters/test_trajectory_converter.py
@@ -15,6 +15,7 @@ from simulariumio.constants import (
     DISPLAY_TYPE,
     DEFAULT_CAMERA_SETTINGS,
     CURRENT_VERSION,
+    MAX_AGENT_ID,
 )
 from simulariumio.exceptions import DataError
 
@@ -54,6 +55,12 @@ def mixed_agents_wrong_display_type1():
 def mixed_agents_wrong_display_type2():
     result = mixed_agents()
     result.agent_data.display_data["Q"].display_type = DISPLAY_TYPE.FIBER
+    return result
+
+
+def mixed_agents_invalid_agent_id():
+    result = mixed_agents()
+    result.agent_data.unique_ids[0][0] = MAX_AGENT_ID + 1
     return result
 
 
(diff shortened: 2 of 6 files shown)


simulariumio/tests/converters/test_trajectory_converter.py
```diff
@@ -57,6 +58,12 @@ def mixed_agents_wrong_display_type2():
     return result
 
 
+def mixed_agents_invalid_agent_id():
+    result = mixed_agents()
+    result.agent_data.unique_ids[0][0] = MAX_AGENT_ID + 1
```
toloudis, line 63: maybe consider defining INVALID_AGENT_ID = MAX_AGENT_ID+1 up above, close to where MAX_AGENT is defined.

simulariumio/constants.py
```diff
@@ -127,3 +127,6 @@ def JMOL_COLORS() -> pd.DataFrame:
     """
     this_dir, _ = os.path.split(__file__)
     return pd.read_csv(os.path.join(this_dir, JMOL_COLORS_CSV_PATH))
+
+
+MAX_AGENT_ID = 0x80000000  # Agent IDs should fit in a 32 bit signed int
```
toloudis, line 132: just curious, why signed int and not unsigned? Also, strategically, is this a check on bad(corrupt or erroneous) data, or a check on scalability to numbers of agents we can't handle? Or something else?

simulariumio/tests/converters/test_trajectory_converter.py
toloudis: oh, nevermind, I just realized this is in a unit test

toloudis (approved):

simulariumio/tests/converters/test_trajectory_converter.py
```diff
@@ -1067,10 +1074,17 @@ def mixed_agents_wrong_display_type2():
...
+            mixed_agents_invalid_agent_id(),
+            {},
+            marks=pytest.mark.raises(exception=DataError),
+        ),
     ],
 )
 def test_trajectory_reader(trajectory, expected_data):
     converter = TrajectoryConverter(trajectory)
     buffer_data = JsonWriter.format_trajectory_data(converter._data)
+    JsonWriter._validate_ids(converter._data)
```
meganrm, line 1088: I don't understand why this is added here

meganrm (approved):
Overall looks good to me, but I would appreciate and answer to my question just for my own understanding

simulariumio/constants.py
blairlyons: unsigned hypothetically would work. And we're checking scalability to make sure all the numbers can be handled by front end

simulariumio/writers/writer.py
```diff
@@ -265,6 +268,17 @@ def _check_agent_ids_are_unique_per_frame(buffer_data: Dict[str, Any]) -> bool:
...
+    @staticmethod
+    def _validate_ids(trajectory_data: TrajectoryData) -> None:
+        """
+        Check if agent unique IDs are valid 32 bit integers
+        returns a message identifying violating agent ID
+        """
+        agent_unique_ids = trajectory_data.agent_data.unique_ids
+        for uid in np.ndarray.flatten(agent_unique_ids):
+            if uid > MAX_AGENT_ID:
+                raise DataError(f"Agent IDs is larger than a 32 bit integer: {uid} ")
```
blairlyons, line 280: Might be useful for the user to know which agent and timestep the bad ID is at, although it would make this more slow to loop through the multi-dimensional array. Maybe there's a way to use slicing?

blairlyons (approved):

PR #117, opened 2022-08-30
import recipeloader from main package

meganrm:
Problem
=======
Simulariumio was importing a class from a directory that moved. 
Additionally, cellpack is now importing a package that is not compatible with python 3.7

Solution
========
To avoid this problem in the future, I changed the import to the package level in cellpack, and am importing the new version here. I also remove python from our tests here since we're building with 3.9

## Type of change

### DIFF
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


ascibisz (approved):

toloudis (approved):
LGTM.  Adding Python 3.10 would probably be good to do soon, also.

blairlyons (approved):

PR #119, opened 2022-09-14
Rewrite Trajectory Converter Tests

ascibisz:
Problem
=======
The trajectory converter tests test huge chunks of data, which makes it difficult to identify what's going on when on e fails.
[Ticket #98](https://github.com/simularium/simulariumio/issues/98)

Solution
========
What I did to solve this problem

Change summary:
---------------
* Broke up the trajectory converter tests into smaller pieces, making it clearer what the issue is when a test fails
* Shortened a couple test files to make tests easier to understand

*Note*: my linear algebra knowledge is limited, so please look closer at `test_mcell_converter` because I tried to write tests to test the functionality of `McellConverter._get_euler_angles()` and some other related functions and they pass, but I certainly could be doing something naive that isn't actually testing sufficiently.


simulariumio/tests/converters/test_trajectory_converter.py
```diff
@@ -59,1011 +62,1180 @@ def mixed_agents_wrong_display_type2():
...
+def test_versions_trajectory(trajectory_version, expected_version):
+    assert trajectory_version == expected_version
+
+
+# test time units
+@pytest.mark.parametrize(
+    "timeUnits, expected_timeUnits",
+    [
+        (
+            default_agents_data["trajectoryInfo"]["timeUnits"],
```
meganrm, line 117: I think the default agents should not include any data that has a default. That way you can test both that the defaults work and that changing the data works. It would also be nice to pull in the default magnitude and default name here

simulariumio/tests/converters/test_trajectory_converter.py
```diff
@@ -59,1011 +62,1180 @@ def mixed_agents_wrong_display_type2():
...
+def test_timeUnits(timeUnits, expected_timeUnits):
+    assert timeUnits == expected_timeUnits
+
+
+# test spatial units
+@pytest.mark.parametrize(
+    "spatialUnits, expected_spatialUnits",
+    [
+        (
+            default_agents_data["trajectoryInfo"]["spatialUnits"],
```
meganrm, line 139: same thing about removing this data from the default agents starting data

simulariumio/tests/converters/test_trajectory_converter.py
```diff
@@ -59,1011 +62,1180 @@ def mixed_agents_wrong_display_type2():
...
+mixed_agents_data = JsonWriter.format_trajectory_data(mixed_agents_converter._data)
+
+# 3 fiber agents with points drawn
+# at given positions for 3 frames, no plots
+fiber_agents_trajectory = fiber_agents()
+fiber_agents_converter = TrajectoryConverter(fiber_agents_trajectory)
+fiber_agents_data = JsonWriter.format_trajectory_data(fiber_agents_converter._data)
+
+
+# test trajectory info
```
meganrm, line 90: After reading through the "test trajectory info" section I think you should make two new sets of data: one that has nothing set in trajectory info that is not required. And one that has everything in the trajectory info set. There is no reason for these to sets of agent data to have anything different about the agents because you're not testing that anyway in these unit tests. They could have the same 1 agent in them with no display data. That way you minimize any change that doesn't have to do with trajectory data breaking these tests.

simulariumio/tests/converters/test_trajectory_converter.py
```diff
@@ -59,1011 +62,1180 @@ def mixed_agents_wrong_display_type2():
...
+
+
+# test model info
+@pytest.mark.parametrize(
+    "modelInfo, expected_modelInfo",
+    [
+        (
+            mixed_agents_data["trajectoryInfo"]["modelInfo"],
+            {
+                "title": "Some agent-based model",
```
meganrm, line 292: same issue here as I wrote above: it's important to test the case where you send in nothing here and make sure it doesn't break

simulariumio/tests/converters/test_trajectory_converter.py
```diff
@@ -59,1011 +62,1180 @@ def mixed_agents_wrong_display_type2():
...
     result = mixed_agents()
     result.agent_data.unique_ids[0][0] = MAX_AGENT_ID + 1
     return result
 
 
+# 3 default agents (radius 5-10) at given positions for 3 frames,
+# test string for plots
+default_agents_trajectory = three_default_agents()
+default_agents_converter = TrajectoryConverter(default_agents_trajectory)
+default_agents_data = JsonWriter.format_trajectory_data(default_agents_converter._data)
```
meganrm, line 75: Another thought, but may be for another PR: 
This is supposed to be testing the `TrajectoryConverter`, but I *think* a lot of the resulting data is handled by the `format_trajectory_data`. While these tests are nice to have, we should have unit tests that actually cover the `TrajectoryConverter` methods without involving any other classes

simulariumio/tests/converters/test_springsalad_converter.py
```diff
@@ -1,575 +1,419 @@
...
+            results["trajectoryInfo"]["size"],
+            {
+                "x": DEFAULT_BOX_SIZE[0],
+                "y": DEFAULT_BOX_SIZE[1],
+                "z": DEFAULT_BOX_SIZE[2],
+            },
+        )
+    ],
+)
+def test_box_size_default(box_size, expected_box_size):
```
meganrm, line 41: I am pretty certain that all these defaults are set by the `TrajectoryConverter`? I think for these package specific converters, we should be testing that it gets data from specific format => trajectory data, I think that's `converter._data`. But we don't need to re-test all this other stuff that we already know works because it works in `TrajectoryConverter`

ascibisz: Box size was weird for me because unlike some other parts of metadata, if nothing is provided, the specific converter has to call `meta_data._set_box_size()` for the default box size to be put in. Of course, for proper unit tests we should be testing `_set_box_size()` independently, but I do want to make sure that the default case doesn't result in breaking behavior in any of the specific converters. That being said, I could pull out the tests for other default tests (like camera position)

meganrm (approved):

simulariumio/tests/converters/test_readdy_converter.py
```diff
@@ -8,262 +8,419 @@
...
-                            "name": "B",
-                            "geometry": {
-                                "displayType": "OBJ",
-                                "url": "c.obj",
-                                "color": "#dfdacd",
-                            },
-                        },
+                },
+                "1": {
+                    "name": "B",
```
blairlyons, line 260: use `name_1` here?

blairlyons (approved):

PR #129, opened 2022-10-19
Updates to smoldyn UI template

blairlyons:
Problem
=======
In the process of implementing autoconversion for Smoldyn, we identified updates that need to be made to the template files:
- we want to be able to validate the file extension for Smoldyn files during autoconversion
- we want to make it easier to convert data from the front end

Solution
========
I added a "match" field for smoldyn_data's file_contents with a regex pattern to validate that the path is for a txt file.

Let me know if there's a better way to do this!

## Type of change

Change summary:
---------------
* add file extension validation for smoldyn file
* also edited some of the names of parameters to make the JSON structure better align with the structure we expect in `FileConverter`, allowing us to better reuse code. This involves:
      * changing parameter names from snake_case to camelCase
      * changing box size's key name from `box_size` to `size`
      * changing the keys for xyz coordinates from {0,1,2} to {x,y,z}

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9]
-        os: [ubuntu-latest, macOS-latest]
+        os: [ubuntu-latest]
 
     steps:
     - uses: actions/checkout@v3
--- a/ui-templates/custom-types/camera_data.json
+++ b/ui-templates/custom-types/camera_data.json
@@ -9,37 +9,37 @@
                 "description": "position of the camera",
                 "required": false,
                 "default": {
-                    "0": 0.0,
-                    "1": 0.0,
-                    "2": 120.0
+                    "x": 0.0,
+                    "y": 0.0,
+                    "z": 120.0
                 },
                 "help": "If not provided, defaults to [0, 0, 120]"
             },
-            "look_at_position": {
+            "lookAtPosition": {
                 "name": "Look at position",
                 "data_type": "coordinates_xyz",
                 "description": "target position to look at",
                 "required": false,
                 "default": {
-                    "0": 0.0,
-                    "1": 0.0,
-                    "2": 0.0
+                    "x": 0.0,
+                    "y": 0.0,
+                    "z": 0.0
                 },
                 "help": "If not provided, defaults to [0, 0, 0]"
             },
-            "up_vector": {
+            "upVector": {
                 "name": "Up vector",
                 "data_type": "coordinates_xyz",
                 "description": "up direction",
                 "required": false,
                 "default": {
-                    "0": 0.0,
-                    "1": 1.0,
-                    "2": 0.0
+                    "x": 0.0,
+                    "y": 1.0,
+                    "z": 0.0
                 },
                 "help": "If not provided, defaults to [0, 1, 0]"
             },
-            "fov_degrees": {
+            "fovDegrees": {
                 "name": "FOV (degrees)",
                 "data_type": "float",
                 "description": "field of view in degrees",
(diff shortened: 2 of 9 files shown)


ui-templates/smoldyn_data.json
```diff
@@ -7,7 +7,8 @@
                 "name": "File",
                 "data_type": "file",
                 "description": "A Smoldyn file",
-                "required": true
+                "required": true,
+                "match": ".*\\\\.txt$"
```
ascibisz, line 11: Does this expression mean that we expect to have "\\" be a part of the path, and if so why? I'm a little confused, but also my regex skills are rusty. Other than that, LGTM!

blairlyons: I have to escape the `\` character for JSON by quadrupling it, but [here's](https://regex101.com/r/B7k5Jj/1) a regex tester that explains what everything is doing otherwise.

meganrm (approved):

blairlyons:
@ascibisz updates lgtm!

.github/workflows/ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9]
-        os: [ubuntu-latest, macOS-latest]
+        os: [ubuntu-latest]
```
meganrm, line 43: why remove this? seems good to have considering we all develop on Macs

ui-templates/custom-types/camera_data.json
```diff
@@ -9,37 +9,37 @@
                 "description": "position of the camera",
                 "required": false,
                 "default": {
-                    "0": 0.0,
-                    "1": 0.0,
-                    "2": 120.0
+                    "x": 0.0,
+                    "y": 0.0,
+                    "z": 120.0
```
meganrm, line 14: is it possible to get these defaults from the data?

ui-templates/custom-types/meta_data.json
```diff
@@ -3,37 +3,37 @@
...
-            "box_size": {
+            "size": {
                 "name": "Box size",
                 "data_type": "coordinates_xyz",
                 "description": "Provide the size of the bounding box in XYZ",
                 "required": false,
                 "help": "If not provided, defaults to [100, 100, 100]"
             },
-            "camera_defaults": {
+            "cameraDefault": {
```
toloudis, line 13: intentional change to remove the trailing `s` ?

ascibisz: good eye, but yes that was intentional. simularium file format just has `cameraDefault`, no `s`

.github/workflows/ci.yml
blairlyons: To be more efficient with test resources, mac is more expensive and since we're all developing on mac it seems ok to skip running CI on it

ui-templates/custom-types/camera_data.json
ascibisz: Honestly, we don't inherently need to define the defaults here, since the autoconversion code takes care of that. Plus, I'm not sure if anything happens with these default values currently, since they aren't pre-populated in the smoldyn form and they don't appear in the json object when the form is submitted... Maybe we should just strip this out?

Also from some quick research, I think bringing the defaults in may be possible, but not simple. I think we'd need to find some sort of json library to use that supports that, regular old json doesn't do variables.

.github/workflows/ci.yml
meganrm: are we running out of action compute time? I thought it was unlimited if you had a public repo. or you just don't like waiting for how long it takes to run all the tests? just curious

toloudis: If we support mac and windows, we probably should run automated tests on mac and windows, especially if binary data is involved, or python packages that are compiled to c++.  This is part of what gh-actions was made for.  Also why not python 3.10 etc?   Also this is out of scope for this PR and has gotten way more attention than it should :)

toloudis (approved):

PR #130, opened 2022-10-19
fix/pin-flake8-pyprojecttoml-to-fix-lint

evamaxfield:
Lint is broken because flake8-pyprojecttoml released a new version (but its breaking?? not really sure).

Context: because projects are slowly moving over to pyproject.toml files to maintain all their metadata, deps, and tooling stuff, we want to store everything we can in that one file. `flake8` the linter doesn't yet support the pyproject.toml file for reading its configurations. the `flake8-pyprojecttoml` library is a sort of experimental plugin that enables that reading.

I will look for better / more stable ways to handle this in the future.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,4 +13,4 @@ repos:
       - id: flake8
         additional_dependencies:
           - flake8-typing-imports>=1.9.0
-          - flake8-pyprojecttoml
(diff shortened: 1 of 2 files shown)


evamaxfield:
Seems like this may be a bit more stable: https://github.com/john-hen/Flake8-pyproject

Will check it out later.

blairlyons (approved):

PR #131, opened 2022-10-27
Fix ReaDDy conversion for types missing display data

blairlyons:
Problem
=======
ReaDDy conversion was failing if display data wasn't provided for every agent type.

Solution
========
Fixed the `ReaddyConverter` so that the default display data that is created actually gets used. Also fixed for the general case in `AgentData`.

## Type of change

Change summary:
---------------
* use default display data in `ReaddyConverter` for any particles without it provided
* create default display data in `AgentData`


ascibisz (approved):

simulariumio/data_objects/agent_data.py
```diff
@@ -536,9 +536,35 @@ def display_type_for_agent(self, time_index: int, agent_index: int) -> DISPLAY_T
...
+        """
+        n_subpoints = self.n_subpoints[time_index][agent_index]
+        default_display_types = [
+            DISPLAY_TYPE.FIBER, 
+            DISPLAY_TYPE.SPHERE_GROUP, 
+            DISPLAY_TYPE.SPHERE,
+        ]
+        for display_type in default_display_types:
+            if n_subpoints % float(SUBPOINT_VALUES_PER_ITEM(display_type)) == 0:
+                return display_type
```
toloudis, line 565: this is potentially a very wrong guess in the future.  there could easily end up being multiple display types with the same number of subpoints per agent.

toloudis, line 565: there is something funky about this indirection..  are you relying on the fact that the subpoint values per item is different for all three?

blairlyons: yeah I am. I could make a constant dict or something with default to use for each values_per_item

toloudis: I agree with that, make it more explicit, or at least just a comment saying "if the number of points per type ever changes then we have to rethink this"

toloudis (approved):

PR #133, opened 2022-11-04
add from_dict to SmoldynData

ascibisz:
Problem
=======
We're implementing autoconversion so that simulariumio can be called from the simularium viewer UI instead of manually using simulariumio to convert files before going to the UI.

After implementing some basic JSON decoding in the autoconversion service, Blair recommended that we do the bulk of that right in simulariumio to remove the potential for redundant code. So, with this PR I implement a method to translate a JSON blob into a `SmoldynData` object, which will be used by Octopus (or perhaps some other service we decide to run autoconversion on)

Kinda an offshoot of [this ticket](https://github.com/simularium/simularium-lambdas/issues/1)

Solution
========
Added `from_dict()` to `SmoldynData`
Used existing `from_dict()` methods from other data objects, although I had to expand the functionality to work for both camel case and snake case
Did a little code cleanup along the way

## Type of change

Open thoughts:
---------------
* I gotta write a test, I just wanna merge my simulariumio test overhaul PR first


simulariumio/smoldyn/smoldyn_data.py
```diff
@@ -71,3 +80,152 @@ def __init__(
             spatial_units if spatial_units is not None else UnitData("m")
         )
         self.plots = plots if plots is not None else []
+
+    @classmethod
+    def from_buffer_data(
```
ascibisz, line 85: Change name to be something more like `from_dict`

ascibisz: and also name the other `from_buffer_data` methods to `from_dict` if they arent actually unpacking a buffer

simulariumio/data_objects/camera_data.py
```diff
@@ -57,38 +58,35 @@ def __init__(
...
-            buffer_data["trajectoryInfo"]["cameraDefault"]
-            if "cameraDefault" in buffer_data["trajectoryInfo"]
-            else None
+    def from_dict(cls, buffer_data: Dict[str, Any]):
+        """
+        Create CameraData object from a simularium JSON dict
+        """
+        # dictionary keys may be in camel case or snake case
+        camera_default = buffer_data.get(
+            "cameraDefault", buffer_data.get("camera_defaults")
```
blairlyons, line 67: probably don't need the `camera_defaults`?

ascibisz: oops yep missed that one

blairlyons: or actually, shouldn't the buffer_data argument be the `cameraDefault` module already?

simulariumio/data_objects/meta_data.py
```diff
@@ -69,23 +70,19 @@ def __init__(
         )
 
     @classmethod
-    def from_buffer_data(cls, buffer_data: Dict[str, Any]):
+    def from_dict(cls, buffer_data: Dict[str, Any]):
```
blairlyons, line 73: rename `buffer_data` to `meta_info` or something?

simulariumio/data_objects/meta_data.py
```diff
@@ -69,23 +70,19 @@ def __init__(
...
             ),
-            camera_defaults=CameraData.from_buffer_data(buffer_data),
-            trajectory_title=buffer_data["trajectoryInfo"]["trajectoryTitle"]
-            if "trajectoryTitle" in buffer_data["trajectoryInfo"]
-            else "",
-            model_meta_data=ModelMetaData.from_buffer_data(buffer_data),
+            camera_defaults=CameraData.from_dict(buffer_data),
+            trajectory_title=buffer_data.get("trajectoryTitle"),
+            model_meta_data=ModelMetaData.from_dict(buffer_data),
+            scale_factor=buffer_data.get("scaleFactor")
```
blairlyons, line 84: should `scaleFactor` have a default?

simulariumio/data_objects/trajectory_data.py
```diff
@@ -74,7 +74,7 @@ def from_buffer_data(
         if display_data is None:
             display_data = {}
         return cls(
-            meta_data=MetaData.from_buffer_data(buffer_data),
+            meta_data=MetaData.from_dict(buffer_data["trajectoryInfo"]),
             agent_data=AgentData.from_buffer_data(buffer_data, display_data),
             time_units=UnitData(
```
blairlyons, line 79: use `UnitData.from_dict(...)`

simulariumio/smoldyn/smoldyn_data.py
```diff
@@ -71,3 +76,53 @@ def __init__(
...
+        """
+        Create SmoldynData from a simularium JSON dict containing buffers
+
+        Parameters
+        ----------
+        buffer_data: Dict[str, Any]
+            JSON dict containing values key value pairs representing the
+            data to be turned into a SmoldynData object
+        """
+        display_data = None
```
blairlyons, line 93: this chunk processing display data could be broken into a function. Likely we'll want to use it for other converters too, so maybe in utils

simulariumio/smoldyn/smoldyn_data.py
```diff
@@ -71,3 +76,53 @@ def __init__(
...
+            time_units = UnitData.from_dict(
+                buffer_data["timeUnits"],
+                default_name="second",
+                default_mag=1.0
+            )
+
+        return cls(
+            meta_data=MetaData.from_dict(buffer_data["metaData"]),
+            smoldyn_file=InputFileData(
+                file_contents=buffer_data["fileContents"]["fileContents"],
```
blairlyons, line 123: is `fileContents` really twice? and I think we should also support `filePath` here

ascibisz: So like would that be a file path to something in an S3 bucket? This will be getting called by octopus, which will be in an EC2 instance, so `filePath` wouldn't really work if the user was just supplying the path to their file locally. I think the UI will need to manage to turn the path (if that's what the user provides) and turn that into actual file contents for autoconversion...

And yeah with the current template, for some reason to get the actual file contents, you have to go to the `fileContents` within `fileContents`

simulariumio/data_objects/trajectory_data.py
ascibisz: ah yes! good eye!

simulariumio/smoldyn/smoldyn_data.py
ascibisz: sounds like a good idea! my quick attempt at doing this just failed, because of where the util file is located and where it's called, I was creating a circular dependency. gotta take a moment to actually plan how the util file should be organized

ascibisz:
TODO: write tests

simulariumio/smoldyn/smoldyn_data.py
blairlyons: yeah we probably won't use file path, definitely not over network, but since this is in simulariumio it could be used for something else, so might as well add that other field. and also let both of them default to empty string

simulariumio/data_objects/display_data.py
```diff
@@ -72,6 +73,25 @@ def __init__(
...
+        display_info: Dict[str, Any],
+        default_display_type: DISPLAY_TYPE = DISPLAY_TYPE.SPHERE
+    ):
+        """
+        Create DisplayData from a simularium JSON dict containing buffers.
+        """
+        if display_info is None:
+            return cls()
+        return cls(
+            name=display_info.get("name"),
```
toloudis, line 88: should name use a default empty string?

simulariumio/data_objects/unit_data.py
```diff
@@ -63,6 +65,23 @@ def multiply(self, multiplier: float):
         self._quantity *= multiplier
         self._update_units()
 
+    @classmethod
+    def from_dict(
+        cls,
+        unit_data: Dict[str, Any],
+        default_name: str = None,
```
toloudis, line 72: just wondering why the choice of None for default_name as opposed to an actual value?

toloudis: I think I see, maybe it's because you don't know if the default unit should be space or time?  I wonder if you can just make it required and not accept a default.

simulariumio/tests/test_from_dict.py
```diff
@@ -0,0 +1,438 @@
...
+            default_smoldyn_dict,
+            SmoldynData(
+                smoldyn_file=InputFileData(file_contents=file_contents_str)
+            ),
+        ),
+    ],
+)
+def test_smoldyn_from_dict(smoldyn_dict, smoldyn_data):
+    result = SmoldynData.from_dict(smoldyn_dict)
+    assert result == smoldyn_data
```
toloudis, line 438: not a huge deal but this file feels like it should be split into many files

toloudis (approved):
LGTM, just a couple questions / optional things to consider.

simulariumio/data_objects/unit_data.py
ascibisz: Yeah so the case I was trying to handle is if someone's filling out the form to convert their data to simularium file format, they don't change the value that's visible in the dropdown for unit name, and thus we'd want to fill that with whatever the form defaulted to showing. So, requiring it would work well for that desired use case / what `SmoldynData` is doing, since I know what I want the default values to be for smoldyn autoconversion.

The problem is, I decided to use `UnitData.fromDict()` in `TrajectoryData.from_buffer_data()` too and I have no idea what to put as the default time units and spatial units there. Perhaps I just shouldn't use it in `trajectory_data.py`? Or I should make up some default values? Open to suggestion, I'm not sure what's best

simulariumio/data_objects/display_data.py
ascibisz: Yeah I'll do that. Hopefully this isn't a case we run into, since name is a required field for display data, but `name=""` does feel better than `name=None`

simulariumio/data_objects/camera_data.py
```diff
@@ -57,38 +58,31 @@ def __init__(
...
     @classmethod
-    def from_buffer_data(cls, buffer_data: Dict[str, Any]):
-        """ """
-        camera_default = (
-            buffer_data["trajectoryInfo"]["cameraDefault"]
-            if "cameraDefault" in buffer_data["trajectoryInfo"]
-            else None
-        )
-        if camera_default is None:
+    def from_dict(cls, camera_default: Dict[str, Any]):
```
blairlyons, line 61: this is probably `camera_info` rather than `camera_default`?

simulariumio/data_objects/display_data.py
```diff
@@ -72,6 +73,25 @@ def __init__(
...
+        display_info: Dict[str, Any],
+        default_display_type: DISPLAY_TYPE = DISPLAY_TYPE.SPHERE
+    ):
+        """
+        Create DisplayData from a simularium JSON dict containing buffers.
+        """
+        if display_info is None:
+            return cls(name="")
+        return cls(
+            name=display_info.get("name", ""),
```
blairlyons, line 88: I like Dan's idea of using "[no name]" as the default, to make it easier for a user to see what happened

simulariumio/data_objects/meta_data.py
```diff
@@ -69,23 +70,20 @@ def __init__(
...
+            box_size=unpack_position_vector(
+                meta_info.get("size"), DEFAULT_BOX_SIZE
             ),
-            camera_defaults=CameraData.from_buffer_data(buffer_data),
-            trajectory_title=buffer_data["trajectoryInfo"]["trajectoryTitle"]
-            if "trajectoryTitle" in buffer_data["trajectoryInfo"]
-            else "",
-            model_meta_data=ModelMetaData.from_buffer_data(buffer_data),
+            camera_defaults=CameraData.from_dict(meta_info.get("cameraDefault")),
+            trajectory_title=meta_info.get("trajectoryTitle"),
```
blairlyons, line 84: should this have a default too?

simulariumio/data_objects/unit_data.py
blairlyons: In `trajectory_data.py` you can use "s" for default time unit and "m" for default spatial unit. Better yet, add those to constants and import them in `trajectory_data.py`

simulariumio/smoldyn/smoldyn_data.py
```diff
@@ -71,3 +79,70 @@ def __init__(
             spatial_units if spatial_units is not None else UnitData("m")
         )
         self.plots = plots if plots is not None else []
+
+    @classmethod
+    def from_dict(
+        cls, buffer_data: Dict[str, Any]
```
blairlyons, line 85: ename `buffer_data` here, maybe `smoldyn_info` to follow the pattern

simulariumio/smoldyn/smoldyn_data.py
```diff
@@ -71,3 +79,70 @@ def __init__(
...
+            time_units = UnitData.from_dict(
+                buffer_data["timeUnits"],
+                default_name="second",
+                default_mag=1.0
+            )
+
+        return cls(
+            meta_data=MetaData.from_dict(buffer_data.get("metaData")),
+            smoldyn_file=InputFileData(
+                file_contents=buffer_data["fileContents"].get("fileContents"),
```
blairlyons, line 130: should these two have defaults?

blairlyons (approved):

simulariumio/smoldyn/smoldyn_data.py
```diff
@@ -71,3 +79,70 @@ def __init__(
             spatial_units if spatial_units is not None else UnitData("m")
         )
         self.plots = plots if plots is not None else []
+
+    @classmethod
+    def from_dict(
+        cls, buffer_data: Dict[str, Any]
+    ):
+        """
+        Create SmoldynData from a simularium JSON dict containing buffers
```
blairlyons, line 88: update this comment too: just "Create SmoldynData from a JSON dict"

simulariumio/smoldyn/smoldyn_data.py
ascibisz: I currently have it set up so `InputFileData.from_dict()` will throw an error if `fileContents` is null, so I think I'd rather leave it without a default. `MetaData.from_dict()` will return a default `MetaData` object if called with null, so I think that's fine too

simulariumio/data_objects/camera_data.py
ascibisz: yeah, you're right :)

Issue #123, opened 2022-09-21
Make agent names case insensitive


blairlyons:
## Use Case
As a user converting data from a supported simulation engine, I want to be able to provide display data for agents without worrying about exactly matching the case of agent names.

## Acceptance Criteria
Agent names in display data mappings can be provided with any combination of lower and upper case characters.

## Details
This applies to data from cellPACK, MCell, MD, ReaDDy, Smoldyn, and SpringSaLaD.

Likely we'll want to lower case strings before comparing them.

PR #135, opened 2022-11-16
Make agent names case insensitive

ascibisz:
Problem
=======
Currently, if you want to provide display data, you have to match the agent name in the data file exactly, including having the proper casing. This is kinda annoying, especially as we move to autoconversion.
[Link to ticket](https://github.com/simularium/simulariumio/issues/123)

Solution
========
Made casing for matching agent names for display data case insensitive 

## Type of change

Change summary:
---------------
* For relevant trajectory converters, made changed the keys in the display data dict, which represent agent names, lowercase
* When comparing those to the agent ids from the file, made sure to cast those to lower case
* Adjusted some of the test files to use agent names with different casing in the display data dict

### DIFF
--- a/simulariumio/md/md_converter.py
+++ b/simulariumio/md/md_converter.py
@@ -62,11 +62,12 @@ class MdConverter(TrajectoryConverter):
         """
         Get the type_name to use for the particle with the given raw type_name
         """
-        if raw_type_name in input_data.display_data:
-            return input_data.display_data[raw_type_name].name
         element_type = guess_atom_element(raw_type_name)
-        if element_type in input_data.display_data:
-            return input_data.display_data[element_type].name + "#" + raw_type_name
+        for name in input_data.display_data:
+            if raw_type_name.lower() == name.lower():
+                return input_data.display_data[name].name
+            if element_type.lower() == name.lower():
+                return input_data.display_data[name].name + "#" + raw_type_name
         return element_type + "#" + raw_type_name
 
     @staticmethod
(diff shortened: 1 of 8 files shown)


toloudis:
Couldn't there be an agent type name "A" and "a" in the same trajectory?  Or even "AA" and "Aa" ?

ascibisz:
> Couldn't there be an agent type name "A" and "a" in the same trajectory? Or even "AA" and "Aa" ?

@toloudis that is a possibility, and basically with this change we're saying if the user provides display data for "aa", we're just going to apply that to all matching agents, regardless of casing. @blairlyons certainly is more familiar than I am, but I think the bad user experience of someone messing up the casing of an agent name during autoconversion and the display data not being applied is pretty significant, and the case of people having agent type names that are only different in casing is pretty rare.

simulariumio/trajectory_converter.py
```diff
@@ -65,16 +65,34 @@ def _get_display_type_name_from_raw(
...
+    @staticmethod
+    def _get_display_data_for_agent(
+        raw_type_name: str,
+        display_data: Dict[str, DisplayData]
+    ) -> DisplayData:
+        """
+        If the provided raw_type_name matches a key in the display data dict,
+        ignoring case, return the corresponding DisplayDat for that key.
+        Otherwise, return None
+        """
```
meganrm, line 90: this description seems like a good set of cases for tests

meganrm (approved):

simulariumio/md/md_converter.py
```diff
@@ -62,29 +62,30 @@ def _get_type_name(raw_type_name: str, input_data: MdData) -> float:
...
-            return input_data.display_data[element_type].name + "#" + raw_type_name
+        for name in input_data.display_data:
+            if raw_type_name.lower() == name.lower():
+                return input_data.display_data[name].name
+            if element_type.lower() == name.lower():
+                return input_data.display_data[name].name + "#" + raw_type_name
         return element_type + "#" + raw_type_name
 
     @staticmethod
     def _get_radius(raw_type_name: str, input_data: MdData) -> float:
```
blairlyons, line 74: not for this PR, but I wonder if this code could be simpler, so we don't have to get display data multiple times for type and radius. It's a bit complicated since these are being called in array functions, but maybe when you're learning more numpy you could think about how to do this better.

blairlyons (approved):

PR #138, opened 2023-02-16
Feature/binary reader leaves binary spatial data

ascibisz:
Problem
=======
For sending binary frame data back from octopus, it'd be great if the SimulariumBinaryReader had an option to leave the spatial data binary encoded when parsing the binary simularium trajectory.

Goes along with [this ticket](https://github.com/simularium/octopus/issues/5)

Solution
========
Added an option to `SimualriumBinaryReader.load_binary()` to leave the trajectory spatial data binary encoded instead of returning spatial data as a list of floats

## Type of change

### DIFF
--- a/simulariumio/readers/simularium_binary_reader.py
+++ b/simulariumio/readers/simularium_binary_reader.py
@@ -155,7 +155,8 @@ class SimulariumBinaryReader:
             frame_n_values = int(frame_lengths[index] / BINARY_SETTINGS.BYTES_PER_VALUE)
             if binary_spatial_data:
                 data = data_as_bytes[
-                    current_frame_offset + 3 : current_frame_offset + frame_n_values
+                    4 * (current_frame_offset + 3) :
+                    4 * (current_frame_offset + frame_n_values)
                 ]
             else:
                 data = list(
--- a/simulariumio/readers/simularium_binary_reader.py
+++ b/simulariumio/readers/simularium_binary_reader.py
@@ -127,7 +127,7 @@ class SimulariumBinaryReader:
         data_as_bytes: np.ndarray,
         data_as_ints: np.ndarray,
         data_as_floats: np.ndarray,
-        binary_spatial_data: bool,
+        parse_data_as_binary: bool,
     ) -> Dict[str, Any]:
         """
         Parse spatial data binary block from a .simularium binary file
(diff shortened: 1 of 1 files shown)


simulariumio/readers/simularium_binary_reader.py
```diff
@@ -151,26 +153,42 @@ def _binary_block_spatial_data(
...
+                    "data": data,
                 }
             )
             current_frame_offset += frame_n_values
         return result
 
     @staticmethod
-    def load_binary(input_file: InputFileData) -> Dict[str, Any]:
+    def load_binary(
+        input_file: InputFileData, binary_spatial_data: bool = False
```
ascibisz, line 180: if anyone has a better name for this bool, let me know! I was having a hard time coming up with a concise variable name for "do the resulting spatial data in binary format or not?"

ascibisz: I guess more `is_binary_spatial_data`? The InputFileData is always binary since this is the binary reader, what I want this bool to represent is whether, as we are parsing the file into individual frames, we return the spatial data as binary or not

meganrm: then `return_spatial_data_as_binary` seems wordy but better.

meganrm: or `spatial_data_as_binary` if you want to get rid of one word

meganrm (approved):

simulariumio/readers/simularium_binary_reader.py
toloudis: how about `parse_spatial_data_as_binary`, or the negative `parse_spatial_data`

simulariumio/readers/simularium_binary_reader.py
```diff
@@ -124,8 +124,10 @@ def _binary_block_json(
     def _binary_block_spatial_data(
         block_index: int,
         block_info: BinaryBlockInfo,
+        data_as_bytes: np.ndarray,
         data_as_ints: np.ndarray,
         data_as_floats: np.ndarray,
+        binary_spatial_data: bool,
```
toloudis, line 130: `data_is_bytes` ? 😄

simulariumio/readers/simularium_binary_reader.py
ascibisz: thanks for the brainstorm, I'm going to go with `parse_spatial_data_as_binary` because it's one character shorter than `return_spatial_data_as_binary` lol

blairlyons (approved):

Issue #63, opened 2021-10-07
Jagged lists for AgentData


blairlyons:
## Use Case
It would be nice if a user with custom data could provide jagged lists to AgentData instead of only numpy arrays, which cannot be jagged.

## Acceptance Criteria
AgentData can be constructed with jagged lists instead of numpy arrays.
- make a classmethod `AgentData.from_lists()` (See `AgentData.from_buffer_data()` for example of this)
- convert lists to numpy arrays, some functions for doing this are already implemented [here](https://github.com/simularium/vivarium-models/blob/main/vivarium_models/processes/simularium_emitter.py), see `SimulariumEmitter.get_simularium_converter()`.

blairlyons:
[this script](https://github.com/simularium/vivarium-models/blob/c44d8ee9899baae7e501f1c82c625bd0da07f6e3/vivarium_models/processes/simularium_emitter.py) has some code that could be useful here

PR #140, opened 2023-03-01
Jagged list for agent data

ascibisz:
Problem
=======
It'd be nice if a user with custom data could provide jagged lists to AgentData instead of only numpy arrays
[Link to ticket](https://github.com/simularium/simulariumio/issues/63)

Solution
========
Added `AgentData.from_lists()`, which takes in a dict of agent data including potentially jagged lists and returns an `AgentData` object. I also added a few tests for this new functionality.

## Type of change


simulariumio/data_objects/agent_data.py
```diff
@@ -330,6 +330,90 @@ def from_buffer_data(
...
+        """
+        Shape a jagged list with 3 dimensions to a numpy array
+        """
+        df = AgentData.fill_df(pd.DataFrame(jagged_3d_list), [0, 0, 0])
+        df_t = df.transpose()
+        exploded = [df_t[col].explode() for col in list(df_t.columns)]
+        return np.array(exploded, dtype=float).reshape((df.shape[0], df.shape[1], 3))
+
+    @staticmethod
+    def get_subpoints_numpy_array(trajectory) -> np.ndarray:
```
blairlyons, line 356: for all of these methods (other than `from_lists()`), they can be private so prepend `_` to the func name, this way there will be less clutter in the auto generated docs

simulariumio/tests/test_agents_from_lists.py
```diff
@@ -0,0 +1,493 @@
+import numpy as np
+
+from simulariumio import AgentData
+
+
+def test_agent_data_jagged_list():
+    list_data = {
```
blairlyons, line 7: It would be nice to define a data class for this instead of using a dict, like `AgentDataLists` or something, so users don't have to guess the key names exactly and so the list dimensions are documented with the types

simulariumio/tests/test_agents_from_lists.py
```diff
@@ -0,0 +1,493 @@
...
+            [0.0, 1.0, 2.0],
+        ],
+        "type_names": [
+            ["H", "A", "C"],
+            ["L", "D", "A", "U"],
+            ["E", "Q", "K"],
+        ],
+        "positions": [
+            [
+                [4.89610492, -29.81564851, 40.77254057],
```
blairlyons, line 27: can some of this data be defined as variables so you don't have to repeat it inline here?

simulariumio/tests/test_agents_from_lists.py
ascibisz: - [ ] create AgentDataList class

blairlyons (approved):
Make sure to add argument types and return types for the new functions in `AgentData`, otherwise looks good

simulariumio/tests/test_agents_from_lists.py
```diff
@@ -0,0 +1,344 @@
...
+    times = [0.0, 1.00001, 2.00001]
+    n_agents = [3, 3, 3]
+    viz_types = [
+        [1001.0, 1001.0, 1001.0],
+        [1001.0, 1001.0, 1001.0],
+        [1001.0, 1001.0, 1001.0],
+    ]
+    unique_ids = [[1.0, 2.0, 3.0], [1.0, 2.0, 3.0], [1.0, 2.0, 3.0]]
+    types = [["H", "A", "C"], ["L", "D", "A"], ["K", "K", "A"]]
+    positions = [
```
meganrm, line 181: where you're just filling in dummy data, could you use `.fill()` so there's just less input data to read?

simulariumio/tests/test_agents_from_lists.py
```diff
@@ -0,0 +1,344 @@
...
+                366.66127719,
+                136.7228888,
+                -210.69313998,
+                -465.59967482,
+                0.0,
+                0.0,
+                0.0,
+                0.0,
+                0.0,
+                0.0,
```
meganrm, line 268: does the length of these zeros matter?

simulariumio/data_objects/agent_data_lists.py
```diff
@@ -0,0 +1,112 @@
...
+        radii: List[List[float]],
+        rotations: List[List[List[float]]] = None,
+        n_subpoints: List[List[int]] = None,
+        subpoints: List[List[List[float]]] = None,
+        display_data: Dict[str, DisplayData] = None,
+        draw_fiber_points: bool = False,
+        n_timesteps: int = -1,
+    ):
+        """
+        This object contains spatial simulation data
```
meganrm, line 45:
```suggestion
        This object contains spatial simulation data as lists
```

simulariumio/tests/test_agents_from_lists.py
```diff
@@ -0,0 +1,344 @@
...
+        times=np.array(times),
+        n_agents=np.array(n_agents),
+        viz_types=np.array(viz_types),
+        unique_ids=np.array(unique_ids),
+        types=types,
+        positions=np.array(positions),
+        radii=np.array(radii),
+        n_subpoints=np.array([[9, 12, 6], [9, 9, 6], [9, 6, 6]]),
+        subpoints=np.array(subpoints),
+    )
```
meganrm, line 341: These tests make me confused about how someone would use this. I assumed from the problem statement that a developer could now send in lists the same way they sent in np arrays and that the code would just handle it. But this looks like the user has to realize they are using lists, which are non standard, and first convert them? Or is this just written out for the tests?

simulariumio/tests/test_agents_from_lists.py
ascibisz: Yes and no. We only care about the non-zero values, those represent the actual subpoint data, but since numpy doesn't support jagged lists, each numpy subpoint array that will eventually be generated for the `AgentData` object will have exactly 12 elements. For this test, I just added various zeros (keeping the number of elements <= 12) to make sure there aren't any weird edge cases when we're filling in the zeros

simulariumio/tests/test_agents_from_lists.py
ascibisz: Ah interesting. I don't see why we can't do it that way, I just started out this route because it was close to the familiar path of `from_buffer_data()` or `from_dict()`. I like that, I'll go work on that now, having either numpy arrays or lists accepted as input parameters and converting everything to numpy arrays on init

- [ ] get rid of AgentDataLists class and handle list input parameters right in AgentData

meganrm (approved):

simulariumio/data_objects/agent_data.py
```diff
@@ -30,31 +30,31 @@
...
     display_data: Dict[str, DisplayData]
     draw_fiber_points: bool
 
     def __init__(
         self,
-        times: np.ndarray,
-        n_agents: np.ndarray,
-        viz_types: np.ndarray,
-        unique_ids: np.ndarray,
+        times: Union[np.ndarray, List[float]],
```
blairlyons, line 48: This is a much better way to implement this, thanks @meganrm 🙏🏼

blairlyons (approved):

Issue #118, opened 2022-09-12
Progress callback


blairlyons:
## Use Case
As a front end developer working on autoconversion, I want to get reports on conversion progress.

## Acceptance Criteria
For any conversion, if given a callback function, call it at a given progress interval, providing the current percent progress.

## Details
For large datasets, especially ones that are being autoconverted over the network with a user waiting at the client website, having a report on progress would be helpful.

For converters that parse text files, determine current progress based on current line out of total lines in input txt file.

PR #142, opened 2023-03-20
Feature/progress callback on converters

ascibisz:
Problem
=======
Add progress callback so octopus can provide more updates on autoconversion progress
[Link to ticket](https://github.com/simularium/simulariumio/issues/118)

Solution
========
Added an optional arguments for a callback function and the number of updates to send to each of the converters. Also added tests.

## Type of change

Open Questions:
----------------
I had difficulties testing for the cellpack converter. I was trying to use the cellpack data from the Jupiter notebook example, since it has multiple ingredients, but I couldn't actually get that example to run without throwing an error. I edited the files down to not include the "snake" ingredient, because the error was thrown in `CellpackConverter._unpack_curve()`, but that's silly I should go in and figure out what's actually going on here.


simulariumio/cellpack/cellpack_converter.py
```diff
@@ -32,7 +33,12 @@
 
 
 class CellpackConverter(TrajectoryConverter):
-    def __init__(self, input_data: CellpackData):
+    def __init__(
+        self,
+        input_data: CellpackData,
+        progress_callback: Callable = None,
+        num_progress_reports: int = 4,
```
toloudis, line 40: is the point of this to limit how many times the progress callback is called?

simulariumio/cellpack/cellpack_converter.py
```diff
@@ -32,7 +33,12 @@
 
 
 class CellpackConverter(TrajectoryConverter):
-    def __init__(self, input_data: CellpackData):
+    def __init__(
+        self,
+        input_data: CellpackData,
+        progress_callback: Callable = None,
```
toloudis, line 39: can this be typed more specifically?  Callable is totally general.  The progress callback seems to take one parameter (the percentage) and is also expected to have a call_count property?

toloudis: or is call_count a property of Mock...?

simulariumio/cellpack/cellpack_converter.py
ascibisz: I have no reason for defaulting to 4 honestly, other than the fact that it's an even number. I could default to 10? I wanted to limit it just so we can configure based on different use cases, I am open to suggestions for different default values though

simulariumio/cellpack/cellpack_converter.py
ascibisz: Call count is a property of Mock, so it's just for testing. I could add a call_count parameter to the callback, but I don't think we need that. I'll certainly go ahead and be more specific for the Callable type

- [ ] make Callable type more specific (ie have 1 float param)

simulariumio/cellpack/cellpack_converter.py
toloudis: 10 somehow feels better to me.  It would be cool if we could do them every 3-5 seconds or something. :wink:

toloudis (approved):

simulariumio/cellpack/cellpack_converter.py
blairlyons: yeah it would be really cool if it could report percent progress at a given time interval

simulariumio/smoldyn/smoldyn_converter.py
```diff
@@ -115,20 +131,22 @@ def _parse_objects(
                     else 1.0
                 )
                 agent_index += 1
+            line_count += 1
+            super().check_report_progress(line_count / len(smoldyn_data_lines))
```
ascibisz, line 135: try doing `self.check_report_progress()`

simulariumio/trajectory_converter.py
```diff
@@ -52,8 +57,28 @@ def __init__(self, input_data: TrajectoryData):
...
+        self.callback_interval = callback_interval
+        self.last_report_time = time.time()
+
+    def check_report_progress(self, percent_complete: float) -> None:
+        current_time = time.time()
+        if (
+            self.progress_callback
+            and current_time > self.last_report_time + self.callback_interval
+        ):
+            self.progress_callback(percent_complete)
```
ascibisz, line 81: missing setting last_report_time here

simulariumio/smoldyn/smoldyn_data.py
```diff
@@ -81,9 +81,7 @@ def __init__(
         self.plots = plots if plots is not None else []
 
     @classmethod
-    def from_dict(
-        cls, smoldyn_info: Dict[str, Any]
-    ):
+    def from_dict(cls, smoldyn_info: Dict[str, Any]):
```
ascibisz, line 84: remove this file

blairlyons (approved):

Issue #146, opened 2023-05-24
Automatically handle scale factor


blairlyons:
## Use Case
As a computational user, I don't want to have to figure out a `scale_factor` value that will look good in the viewer in order for my data to not clip or have weird shading when I first load it.

## Acceptance Criteria
When a user provides spatial values that are very large or very small, that is detected during conversion and multiplied by an additional scale factor that will result in values in the range the viewer expects.

## Details
During conversion, find the min and max X, Y, and Z values as we go through the data for each timestep and agent. Once we've looped through all the data, calculate the largest dimension of the range of the data (`np.amax(max - min)`) and make sure it is between the min and max spatial values that work for the viewer, if not, multiply everything by an additional scale factor to get it in this range. 

The additional scale factor needs to be applied to all spatial values (positions, fiber subpoints, box size) and spatial units, so that the UI still displays correct spatial values with correct units, but internally the viewer can use the agent positions etc that are in a better range for rendering.

Part of this task is figuring out what are the min and max spatial values for the viewer. From testing, we'd expect it to be around 10 - 100 but would be good to verify this by trying to load trajectories with different magnitudes of spatial values (i.e. with values near +/- 0.01, 0.1, 1, 10, 100, 1000, or 10000) and checking camera clipping and shading are reasonable.

PR #148, opened 2023-06-07
Feature/auto handle scale factor

ascibisz:
Problem
=======
It's confusing for users to have to come up with an appropriate scale factor that'll look good in the simularium viewer. Instead, simulariumio should be able to choose an appropriate scale factor based on the spatial positions.
[Link to ticket](https://github.com/simularium/simulariumio/issues/146)

Solution
========
Based on manually trying on different scale factors and seeing what positional ranges looked bad, I decided the viewer's ideal range is 5-70, so I made a new constant for that. In each of the converters, I kept track of the ranges of the XYZ coordinates when spatial data is being processed, and then come up with a scale factor accordingly. I then scale radii, positions, subpoints, and box size accordingly. If a scale factor was provided, that will still be used instead of the calculated one, to maintain previous functionality.

## Type of change


simulariumio/cytosim/cytosim_converter.py
```diff
@@ -235,13 +236,20 @@ def _parse_objects(
...
                         float(columns[3].strip("+,")),
                     ]
                 )
+                TrajectoryConverter.check_max_min_coordinates(
+                    max_dimensions,
+                    min_dimensions,
+                    result.subpoints[time_index][agent_index][
+                        subpoint_index : subpoint_index + 3
+                    ],
+                )
```
blairlyons, line 252: Instead of doing this each time we read an XYZ position, we can do it after we finish going through and getting all the positions using numpy slicing. Let's talk about it in 1:1

simulariumio/cytosim/cytosim_converter.py
```diff
@@ -260,20 +267,30 @@ def _parse_objects(
...
+                TrajectoryConverter.check_max_min_coordinates(
+                    max_dimensions,
+                    min_dimensions,
+                    result.positions[time_index][int(result.n_agents[time_index])],
                 )
                 result.n_agents[time_index] += 1
             self.check_report_progress(overall_line / total_lines)
 
+        if scale_factor is None:
```
blairlyons, line 285: We can use numpy slicing to get dimensions here

simulariumio/trajectory_converter.py
```diff
@@ -81,6 +81,40 @@ def check_report_progress(self, percent_complete: float) -> None:
...
+        min_dimensions: np.array,
+    ) -> float:
+        """
+        Return a scale factor, using the given min and max
+        dimensions, so that the final range of agent locations
+        is within the dimensions defined by VIEWER_DIMENSION_RANGE
+        """
+        range = max(max_dimensions - min_dimensions)
+        scale_factor = 1
+        if range == 0:
```
blairlyons, line 110: It makes sense to check this so you don't divide by zero, but since it's a float it's unlikely to be exactly zero, so this line seems weird to me. You could use `numpy.isclose()` but also maybe it doesn't matter since division will work fine if it's slightly off zero.

simulariumio/cytosim/cytosim_converter.py
blairlyons: `x_max =  max(pos[:,0])`

simulariumio/trajectory_converter.py
```diff
@@ -81,6 +81,93 @@ def check_report_progress(self, percent_complete: float) -> None:
             self.progress_callback(percent_complete)
             self.last_report_time = current_time
 
+    def _filter_agent_data(data: np.array, n_agents: np.array):
```
ascibisz, line 84: add comment

ascibisz: better name: get_real_agent_positions or something

simulariumio/constants.py
```diff
@@ -170,3 +170,8 @@ def JMOL_COLORS() -> pd.DataFrame:
 ]
 
 MAX_AGENT_ID = 0x80000000  # Agent IDs should fit in a 32 bit signed int
+
+
+class VIEWER_DIMENSION_RANGE:
+    MIN = 5
+    MAX = 70
```
toloudis, line 177: as a software engineer I would be more comfortable using "4" (or "8") and "64" but it probably makes no difference.  I wonder what kind of scene would break this?  A very very densely packed one , I guess?

simulariumio/smoldyn/smoldyn_converter.py
```diff
@@ -137,7 +134,18 @@ def _parse_objects(
...
+
+        if input_data.meta_data.scale_factor is None:
+            # If scale factor wasn't provided, calculate one
+            scale_factor = TrajectoryConverter.calculate_scale_factor(
+                result.positions, result.radii, result.n_agents, is_2D=(not is_3D)
+            )
+        else:
+            scale_factor = input_data.meta_data.scale_factor
+        result.radii = scale_factor * result.radii
+        result.positions = scale_factor * result.positions
```
toloudis, line 146: is the above 10 lines or so repeated in all the other converters? possible DRY refactor?

simulariumio/trajectory_converter.py
toloudis: I would agree with picking an epsilon here!

toloudis (approved):

simulariumio/md/md_converter.py
```diff
@@ -198,17 +196,33 @@ def _read_universe(self, input_data: MdData) -> AgentData:
...
+                result.positions - result.radii[:, :, np.newaxis], result.n_agents
+            )
+            scale_factor = TrajectoryConverter.calculate_scale_factor(
+                max_dimensions, min_dimensions
+            )
+        else:
+            scale_factor = input_data.meta_data.scale_factor
+        result.radii = scale_factor * result.radii
+        result.positions = scale_factor * result.positions
+        return result, scale_factor
```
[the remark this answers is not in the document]
blairlyons: agree with this, would be great to consolidate some code

simulariumio/trajectory_converter.py
```diff
@@ -81,6 +82,139 @@ def check_report_progress(self, percent_complete: float) -> None:
...
+        y_data = np.array([])
+        z_data = np.array([])
+        x = data[:, :, 0]
+        y = data[:, :, 1]
+        z = data[:, :, 2]
+        for i in range(len(n_agents)):
+            x_data = np.append(x_data, x[i][0 : int(n_agents[i])])
+            y_data = np.append(y_data, y[i][0 : int(n_agents[i])])
+            z_data = np.append(z_data, z[i][0 : int(n_agents[i])])
+        return (x_data, y_data, z_data)
```
blairlyons, line 102: return a numpy array here?

simulariumio/cellpack/cellpack_converter.py
```diff
@@ -276,18 +267,17 @@ def _process_ingredients(
         self,
         all_ingredients,
         time_step_index: int,
-        scale_factor: float,
         box_center: np.array,
+        scale_factor: Union[float, None],
```
blairlyons, line 271: could use Optional here instead of Union?

simulariumio/data_objects/meta_data.py
```diff
@@ -30,7 +30,7 @@ def __init__(
         self,
         box_size: np.ndarray = None,
         camera_defaults: CameraData = None,
-        scale_factor: float = 1.0,
+        scale_factor: Union[float, None] = None,
```
blairlyons, line 33: could also be Optional instead of Union

simulariumio/physicell/physicell_converter.py
```diff
@@ -316,20 +312,30 @@ def _get_trajectory_data(
...
                 agent_index += 1
             result.n_agents[time_index] = agent_index
+
+        if input_data.meta_data.scale_factor is None:
+            # If scale factor wasn't provided, calculate one
+            scale_factor = TrajectoryConverter.calculate_scale_factor(result)
+        else:
+            scale_factor = input_data.meta_data.scale_factor
+        for index in range(dimensions.total_steps):
+            # do not scale radii for subcells
```
blairlyons, line 327: why don't we want to scale subcell radii?

simulariumio/tests/converters/test_cellpack_converter.py
```diff
@@ -27,6 +29,8 @@
 
 converter = CellpackConverter(data)
 results = JsonWriter.format_trajectory_data(converter._data)
+range = 200.0
+auto_scale_factor = VIEWER_DIMENSION_RANGE.MAX / range
```
blairlyons, line 33: what is the auto scale factor?

simulariumio/tests/converters/test_cytosim_converter.py
```diff
@@ -115,13 +116,13 @@ def test_box_size_provided(box_size, expected_box_size):
     assert box_size == expected_box_size
 
 
+auto_scale_factor = VIEWER_DIMENSION_RANGE.MIN / (0.400052 + 0.001)
```
blairlyons, line 119: where do these numbers come from?

simulariumio/tests/converters/test_cellpack_converter.py
ascibisz: that's the scale factor that's automatically generated. I didn't spend much time thinking up a name for it since it's just in a test, but I'm open to suggestion

simulariumio/tests/converters/test_cytosim_converter.py
ascibisz: they're pulled from the test data, that represents the max range of the positional data, which determines the scale factor. I can pull that out and call it the range, so it's a little more clear?

- [ ] make range variable

simulariumio/physicell/physicell_converter.py
ascibisz: oh goodness I don't remember, I think this came out of when I manually checked that all of the scaled results looked like what I would expect them to be. I'll double check, and if there is a good reason I'll make the comment more descriptive

- [ ] figure out why "do not scale radii for subcells"

ascibisz: ok so I lied, I came across this because this matches the previous functionality so if i just did `radii *= scale_factor` for all radii, a previous test would fail. That's because "subcells" are represented as sphere groups, and so previously if someone had provided a scale factor, we wouldn't apply that to the agent of the sphere group's radii. I imagine it doesn't actually matter which way we do it, because I don't think the sphere group's agent radii matters? since the radii for each of the spheres of the sphere group is in the subpoints?

blairlyons (approved):

Issue #145, opened 2023-05-24
Center Smoldyn and MEDYAN data


blairlyons:
## Use Case
As a computational user with data from Smoldyn or MEDYAN, I want to be able to center my spatial data as a part of the conversion process so that I don't have to filter it in a separate step after converting.

## Acceptance Criteria
`SmoldynData` and `MedyanData` have a new `center` bool member that defaults to `True` and if `True`, the spatial values of the data are centered around the origin (0, 0, 0) during conversion.

## Details
The examples notebooks for [Smoldyn](https://github.com/simularium/simulariumio/blob/main/examples/Tutorial_smoldyn.ipynb) and [MEDYAN](https://github.com/simularium/simulariumio/blob/main/examples/Tutorial_medyan.ipynb) include an extra step for filtering with the `TranslateFilter` after conversion. We want to elimate the need for this step.

The [implementation](https://github.com/simularium/simulariumio/blob/main/simulariumio/filters/translate_filter.py) of `TranslateFilter` shows basically how we want to do this, except there will be one `translation` value that is used for every agent, instead of different translations for different agent IDs.

In order to get the translation amount, after we parse through the data, we can get the min and max values for X, Y, and Z using numpy slicing, and then offset all the positions by `-0.5 * (max - min)`.

PR #152, opened 2023-07-06
Center position data for smoldyn and medyan

ascibisz:
Problem
=======
It's annoying that for smoldyn and medyan data, centering the data is a separate translation filter step after converting the trajectory to simularium file format. We want to be able to center spatial data as a part of the conversion process instead of requiring an extra step.
[Link to ticket](https://github.com/simularium/simulariumio/issues/145)

This change adds an optional boolean parameter `center` to `SmoldynData` and `MedyanData` which defaults to true, and if true then the spatial data will be translated to be centered around the origin (0, 0, 0) during conversion. If false, the spatial data will not be translated.

## Type of change

### DIFF
--- a/simulariumio/trajectory_converter.py
+++ b/simulariumio/trajectory_converter.py
@@ -233,6 +233,7 @@ class TrajectoryConverter:
         min_dimensions: np.array,
         max_dimensions: np.array,
     ) -> float:
+        range = max(max_dimensions - min_dimensions)
         scale_factor = 1
         if np.isclose(range, 0):
             return scale_factor
--- a/simulariumio/trajectory_converter.py
+++ b/simulariumio/trajectory_converter.py
@@ -171,16 +171,20 @@ class TrajectoryConverter:
         max_subpoints = int(np.amax(data.n_subpoints))
         for time_index in range(total_steps):
             for agent_index in range(int(data.n_agents[time_index])):
-                data.positions[time_index][agent_index] += translation
                 display_type = data.display_type_for_agent(time_index, agent_index)
-                if max_subpoints > 0 and display_type == DISPLAY_TYPE.FIBER:
+                has_fiber_subpoints = (
+                    max_subpoints > 0 and display_type == DISPLAY_TYPE.FIBER
+                )
+                if has_fiber_subpoints:
                     # only translate subpoints for fibers, since sphere group
                     # subpoint positions are relative to agent's position, and no
                     # other display types have subpoints currently
                     sp_items = Filter.get_items_from_subpoints(
                         data, time_index, agent_index
                     )
-                    if sp_items is not None:
+                    if sp_items is None:
+                        has_fiber_subpoints = False
+                    else:
                         # translate subpoints for fibers
                         n_items = sp_items.shape[0]
                         for item_index in range(n_items):
@@ -189,6 +193,13 @@ class TrajectoryConverter:
                         data.subpoints[time_index][agent_index][
                             :n_sp
                         ] = sp_items.reshape(n_sp)
+                if not has_fiber_subpoints:
+                    # agents for fibers don't have their own position data, so only
+                    # translate agents without fiber subpoints. eventually, if fiber
+                    # subpoints are relative to agent position, we can just translate
+                    # the agent position and leave the subpoints as is
+                    data.positions[time_index][agent_index] += translation
+
         return data
 
     @staticmethod
--- a/simulariumio/trajectory_converter.py
+++ b/simulariumio/trajectory_converter.py
@@ -171,20 +171,16 @@ class TrajectoryConverter:
         max_subpoints = int(np.amax(data.n_subpoints))
         for time_index in range(total_steps):
             for agent_index in range(int(data.n_agents[time_index])):
+                data.positions[time_index][agent_index] += translation
                 display_type = data.display_type_for_agent(time_index, agent_index)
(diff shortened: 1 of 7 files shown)


simulariumio/trajectory_converter.py
```diff
@@ -148,6 +149,35 @@ def get_subpoints_xyz(subpoints: np.array, n_subpoints: np.array) -> np.array:
...
+    @staticmethod
+    def translate_positions(data: AgentData, translation: np.ndarray) -> AgentData:
+        total_steps = data.times.size
+        max_subpoints = int(np.amax(data.n_subpoints))
+        for time_index in range(total_steps):
+            for agent_index in range(int(data.n_agents[time_index])):
+                display_type = data.display_type_for_agent(time_index, agent_index)
+                translate_subpoints = (
+                    max_subpoints > 0 and display_type != DISPLAY_TYPE.SPHERE_GROUP
+                )
```
meganrm, line 161: is this just checking if the agent has subpoints? If so, I think it would be better to name `translate_subpoints` something like `has_subpoints` because at first glance I thought this was an option and I was confused why you would sometimes not want to translate the subpoints, and like 177-178 is pretty confusing

simulariumio/trajectory_converter.py
```diff
@@ -148,6 +149,35 @@ def get_subpoints_xyz(subpoints: np.array, n_subpoints: np.array) -> np.array:
...
+                        # translate subpoints for fibers
+                        n_items = sp_items.shape[0]
+                        for item_index in range(n_items):
+                            sp_items[item_index][:VALUES_PER_3D_POINT] += translation
+                        n_sp = int(data.n_subpoints[time_index][agent_index])
+                        data.subpoints[time_index][agent_index][
+                            :n_sp
+                        ] = sp_items.reshape(n_sp)
+                if not translate_subpoints:
+                    data.positions[time_index][agent_index] += translation
```
meganrm, line 178: I'm also confused why you wouldn't do this line no matter what if they have subpoints. I'm sure it's working, I think it could just use more comments and explanations

blairlyons: I can't remember exactly, but I don't think the viewer handles local position offsets well for fibers, so we leave positions at zero and only use subpoint positions. @ascibisz it would be good to test the current viewer with fibers with non-zero positions to check this is still the case, we can chat about it tomorrow.

blairlyons (approved):

simulariumio/trajectory_converter.py
toloudis: Please put an issue in simularium-viewer if it is true that we can't position fibers in this way.

ascibisz: Ok so I played around in the simularium viewer and the viewer just ignores the agent data position for fibers! So the if statement here _can_ be removed, even though we technically don't need to move the agent position, it doesn't really matter if we do. I think it's less confusing if we just translate it anyways, so I'll do that

ascibisz: - [ ] remove if statement for translating agent position

toloudis: Just to make sure I understand, if the agent position for a fiber is not (0,0,0), then shouldn't the subpoints be relative to that position?

toloudis: One more thought to add:  if the data is generated so that every fiber is at position 0,0,0 and rotation 0,0,0, and the subpoints are baked to the right positions, the viewer should work and people can continue making data like that.  I'm just saying that the fact that we don't account for position and rotation in the viewer should be an issue ticket!  I want to revisit that and make sure we fully understand the technical limitation and fix it if we can.

ascibisz: Maybe we should discuss this more during a dev chat, since I really haven't dug very much into the viewer code to know how all of this works.  What I do know is that when we generate data for fibers, the agent position is always (0, 0, 0) and all of the actual position information is contained in the subpoints. So I guess technically the subpoints are relative to the agent position, in the same way that any position data is relative to the origin! But yeah if you change the agent position and leave the subpoints the same, it will look the same. I haven't played around with rotations much, but I'll do that now

Also for sphere groups, the subpoint positions are relative to the agent position though, so that's fun.

toloudis: I'm sure you're right but I just think we should have the data output in a way that is at least consistent with being able to rotate and translate any agent.(So keep the center at 0,0,0)  I'm not going to suggest changing the way we output fibers until we know how hard it will be to make the viewer handle it!  Also, if the idea of having a center and rotation for a fiber is a concept most engines don't deal with anyway, then it could be a moot point.

ascibisz: Ok @blairlyons let's talk about this during 1:1 today, but I'll probably just revert the commit where I removed the if statement that kept the agent position for fibers at (0, 0, 0)

ascibisz: alright @toloudis I switched back to keeping the agent position at the origin when we translate fiber agents that have subpoints, because at least that's consistent with how we've historically translated agents.

I also cut a [ticket](https://github.com/simularium/simularium-viewer/issues/332) for simularium-viewer to look into this more and likely switch to interpreting fiber subpoints as relative to the agent position, like we do for sphere groups!

toloudis (approved):
LGTM.  I will try to look at the translation/rotation for fibers in the viewer soon and get that resolved (or have a reason why we can't...)

Issue #149, opened 2023-06-21
Update example Jupyter notebooks


blairlyons:
## What needs to happen?
Some of the Jupyter notebook examples are out of date from recent changes. We need to test all the notebooks and fix any errors or out of date instructions.

## Why should we do this?
It's important for new users to be able to easily find and run working up to date examples.

## When does this need to get done?
Summer 2023

## Potential Reviewers
(while Blair is away)
@meganrm @toloudis @jessicasyu

PR #154, opened 2023-07-17
Fix/update examples

ascibisz:
Problem
=======
The Jupyter notebook examples for simularium io are not up to date with the current state of simularium io, and as a result some of the examples are raising errors when they're run!  We want our examples to work, so this ticket was to debug the issues and update the Jupyter notebooks accordingly.
[Link to ticket](https://github.com/simularium/simulariumio/issues/149)

Solution
========
Mostly, the changes include:
* Adding `DisplayType` field in for `DisplayData` since that is now a required parameter
* Fixing a bug I created with part of the CellPack converter I created in a recent PR 😭 but missed because I hadn't tested with any curve data
* Changing the Medyan and Smoldyn examples to use the automatic centering functionality that's being added
* Fixing a few other minor bugs

## Type of change


examples/Tutorial_smoldyn.ipynb
```diff
@@ -120,9 +122,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Once your data is shaped like in the `example_data` object, you can use the converter to generate the file at the given path:\n",
-    "\n",
-    "(since this model's coordinates are all positive, use a `TranslateFilter` to center the data in the viewer.)"
```
ShrimpCryptid, line 125: Is this behavior now automatically handled as part of the conversion API?

ShrimpCryptid (approved):
Looks good to me! I'm not super familiar with the Simularium repo so I'll defer to any other opinions :)

examples/Tutorial_smoldyn.ipynb
ShrimpCryptid: OOPS found it in the description, nvm!

interim17:
I'm also not sure protocol for testing simio branches, but this overall looks good to me too.

interim17 (approved):

meganrm:
Nice work. is there any way to have the Jupyter notebook examples auto tested with we run our other tests?

ascibisz:
> Nice work. is there any way to have the Jupyter notebook examples auto tested with we run our other tests?

Good idea! I just used a simple package that'll make sure the Jupyter notebooks execute without raising any errors, and added that to `just test`. Obviously it won't look at the outputted files, but this'll keep us from getting to the state where we're raising errors in the example notebooks again

meganrm (approved):

Issue #153, opened 2023-07-10
Issues with appending agents to TrajectoryData object


mogres:
## System and Software
* simulariumio Version: 1.7.1
* Python Version: 3.10.10
* Operating System: linux

## Description
While appending new `AgentData` obtained from the `CytosimConverter` to an existing `TrajectoryData` object, I ran into the following issues:

1. `used_uids`, `new_agent_index` and `raw_uid` are being saved as floats instead of ints, which caused issues during indexing
2. The `result.display_data` dictionary was not updated during the call to `append_agents`, which resulted in the added agents defaulting to a sphere geometry
3. the default buffer size used by `get_copy_with_increased_buffer_size` was too large (1000) for my use case. This resulted in 1000 elements being added every step even when they were not necessary.

## Expected Behavior
New `AgentData` should be added to the existing `TrajectoryData` with its `DisplayData` carried over. 

## Reproduction
1. Create a `CytosimData` object (`fiber_data`) from a cytosim output file
2. Create a `CytosimConverter` object (`cytosim_converter`) from the `CytosimData` object.
3. Repeat steps 1 and 2 with a different output file to get a new `CytosimConverter` object `new_cytosim_converter`
4. Append new agent data to the original using:
```
trajectory_data = cytosim_converter._data  # this is a TrajectoryData object
new_agent_data = new_cytosim_converter._data.agent_data  # this is an AgentData object
trajectory_data.append_agents(new_agent_data)
```
I can share the data and a notebook with the reproduction example if needed.

## More Information
Local changes I made to fix these issues are available at https://github.com/mogres/simulariumio/tree/multiple_cytosim_repeats

blairlyons:
@mogres is there still work to do on this issue or was it closed by https://github.com/simularium/simulariumio/pull/155

mogres:
> @mogres is there still work to do on this issue or was it closed by #155

Yes, we can close this issue now.

PR #155, opened 2023-08-16
Append multiple agents to TrajectoryData objects

mogres:
Problem
=======
Refer to #153 

Solution
========
* updated buffer size for single agents
* enforce integer datatypes for indices

## Type of change

Steps to Verify:
----------------
Refer to this [notebook](https://github.com/simularium/subcell-analysis/blob/simularium_multiple_repeats/notebooks/create_simularium_outputs.ipynb) which creates a single simularium file from multiple cytosim simulations

Screenshots:
-----------------------
Simularium trajectory with multiple actin filaments visualized simultaneously
image

Keyfiles:
-----------------------
`simulariumio/data_objects/trajectory_data.py`

### DIFF
--- a/simulariumio/data_objects/trajectory_data.py
+++ b/simulariumio/data_objects/trajectory_data.py
@@ -11,6 +11,7 @@ from .agent_data import AgentData
 from .unit_data import UnitData
 from .meta_data import MetaData
 from .display_data import DisplayData
+from .dimension_data import DimensionData
 
 ###############################################################################
 
@@ -18,6 +19,11 @@ log = logging.getLogger(__name__)
 
 ###############################################################################
 
+BUFFER_SIZE_INC: DimensionData = DimensionData(
+            total_steps=1,
+            max_agents=1,
+            max_subpoints=1,
+        )
 
 class TrajectoryData:
     meta_data: MetaData
@@ -95,7 +101,7 @@ class TrajectoryData:
         added_dimensions = new_agents.get_dimensions()
         new_dimensions = current_dimensions.add(added_dimensions, axis=1)
         result = self.agent_data.check_increase_buffer_size(
-            new_dimensions.max_agents - 1, axis=1
+            new_dimensions.max_agents - 1, axis=1, buffer_size_inc=BUFFER_SIZE_INC
         )
         # add new agents
         result.n_agents = np.add(result.n_agents, new_agents.n_agents)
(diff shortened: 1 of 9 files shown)


ascibisz:
Looks good to me!

ascibisz (approved):

ShrimpCryptid (approved):
This looks reasonable to me! I'm not super familiar with Simularium's codebase so feel free to take this with a grain of salt :)

toloudis:
I see broken tests.. floating point comparisons.. and combined with the int casting in the code change, I'm not sure what happened.  But the precision for the float comparisons seems way too high.  Usually in a unit test you would never test for equality with floating point numbers that possibly had math run on them, but use an epsilon.

mogres:
> I see broken tests.. floating point comparisons.. and combined with the int casting in the code change, I'm not sure what happened. But the precision for the float comparisons seems way too high. Usually in a unit test you would never test for equality with floating point numbers that possibly had math run on them, but use an epsilon.

Yup, looks like very small floating point errors are causing the tests to fail in an OS-dependent way. I'm not sure what the best way to deal with this is, would be happy to get some input. Any thoughts @ascibisz?

ShrimpCryptid:
Looked into the bug a bit, currently there's an equality assertion between two tuples containing a dictionary and a float array in a few tests. Pytest does have an `approx` function that handles float array comparison with an epsilon (https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-approx) that might work if we compare the elements of the tuple?

Something like:
```
# test_cellpack_convert.py, 156
assert expected_bundleData_data[0] == bundleData["data"][0] && expected_bundleData_data[1] == pytest.approx(bundleData["data"][1]
```

Also relevant stackoverflow thread if this is happening elsewhere: https://stackoverflow.com/questions/56046524/check-if-python-dictionaries-are-equal-allowing-small-difference-for-floats

mogres:
> Looked into the bug a bit, currently there's an equality assertion between two tuples containing a dictionary and a float array in a few tests. Pytest does have an `approx` function that handles float array comparison with an epsilon (https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-approx) that might work if we compare the elements of the tuple?
> 
> Something like:
> 
> ```
> # test_cellpack_convert.py, 156
> assert expected_bundleData_data[0] == bundleData["data"][0] && expected_bundleData_data[1] == pytest.approx(bundleData["data"][1]
> ```
> 
> Also relevant stackoverflow thread if this is happening elsewhere: https://stackoverflow.com/questions/56046524/check-if-python-dictionaries-are-equal-allowing-small-difference-for-floats

I'm trying out a fix using numpy isclose which might work. Will keep you posted!

Issue #151, opened 2023-06-28
Add New Data Types for Simularium Frame Data


ascibisz:
## Use Case
As an octopus developer, I would like to have data types to represent simularium file data, which should live in simulariumio. Although there are some differences between how binary and json simularium data is stored, there should be a universal set of functions that developers can call to get specific frames, the number of frames, overall file contents, etc.

## Acceptance Criteria
Build classes in SimulariumIO that are described in this [miro board](https://miro.com/app/board/uXjVPnOF5p4=/), under "New SimulariumIO Data Types (pseudo code)"

Add tests for these new data objects

Push a new release of simulariumio

PR #157, opened 2023-08-30
Add New Data Types for Simularium Frame Data

ascibisz:
Problem
=======
It'd be nice to have simularium data objects that are designed for providing frame data for a trajectory, so that the octopus repo doesn't have to define it's own. Should match what is defined in this [miro board](https://miro.com/app/board/uXjVPnOF5p4=/) under "New SimulariumIO Data Types (pseudo code)"

[Ticket](https://github.com/simularium/simulariumio/issues/151)

There will be a forthcoming PR for using these new data objects in Octopus

## Type of change


simulariumio/data_objects/simularium_file_data.py
```diff
@@ -0,0 +1,42 @@
+from typing import Dict, Union
+from .trajectory_data import TrajectoryData
+from .frame_data import FrameData
+
+
+class SimulariumFileData:
```
toloudis, line 6: Since this looks like the same API I set up in the viewer, I think it might be important to note that this api is good for reading but useless for writing.   
Why does the constructor take both a file name and a file contents?  Are they mutually exclusive?

simulariumio/data_objects/frame_data.py
```diff
@@ -0,0 +1,79 @@
...
+            Spatial data for the frame, as a byte array for binary encoded
+            .simularium files or as a dict for JSON .simularium files
+        """
+        self.frame_number = frame_number
+        self.n_agents = n_agents
+        self.time = time
+        self.data = data
+
+
+class DataIndices:
```
toloudis, line 29: only used by binary_data right?   So it could live entirely in there.  Same for FrameMetadata?

simulariumio/data_objects/binary_data.py
```diff
@@ -0,0 +1,131 @@
...
+            + BINARY_SETTINGS.BLOCK_HEADER_N_VALUES
+        )
+        n_frames = self.file_data.int_view[spatial_block_offset + 1]
+        current_frame_offset = spatial_block_offset + 2 + 2 * n_frames
+        for i in range(n_frames):
+            offset = (
+                self.file_data.int_view[spatial_block_offset + 2 + (2 * i)]
+                + spatial_indices.offset
+            )
+            length = self.file_data.int_view[spatial_block_offset + 3 + 2 * i]
```
toloudis, line 64: some of these "magic number" offsets of 2 or 3 could use either comments or symbolic constants.

simulariumio/data_objects/binary_data.py
```diff
@@ -0,0 +1,131 @@
...
+            self.file_data.byte_view
+        )
+        for block_index in range(block_info.n_blocks):
+            block_type_id = SimulariumBinaryReader._binary_block_type(
+                block_index, block_info, self.file_data.int_view
+            )
+            self.block_indices[block_type_id] = DataIndices(
+                offset=block_info.block_offsets[block_index],
+                length=block_info.block_lengths[block_index],
+            )
```
toloudis, line 47: the above code is pretty close to what BinaryBlockInfo is doing too.. maybe they could be combined somehow.  I don't think they are identical but kinda similarly collecting up block offsets and lengths.

simulariumio/data_objects/simularium_file_data.py
ascibisz: I don't think the file name is actually necessary. I had thought about removing it before, but evidently I did not. I'm just going to remove that

- [ ] remove file_name param

simulariumio/data_objects/frame_data.py
ascibisz: - [ ] move data classes that are only relevant for binary_data to binary_data

simulariumio/data_objects/json_data.py
```diff
@@ -0,0 +1,74 @@
...
+        Parameters
+        ----------
+        file_name : str
+            Name of the file
+        file_contents : str
+            A string of the data of an open .simularium file
+        """
+        self.file_name = file_name
+        self.data = json.loads(file_contents)
+        self.n_agents = AgentData.from_buffer_data(self.data).n_agents
```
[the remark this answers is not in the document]
ascibisz: Yeah it's a fairly heavy operation. Where would you propose caching it? This is one of the weird situations where in the binary encoding, we save the number of agents in each frame explicitly, but in the json encoding, we don't. And that matters when we're trying to build frames in binary encoding, which need to have n_agents, for a json encoded simularium trajectory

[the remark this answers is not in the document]
toloudis: I don't want to belabor this if most of our files are not going to be in json format anyway.  I was sort of wondering if the results of AgentData.from_buffer_data would make sense to just live here as a member variable.  Like would it be useful to stash it here for any other operations?   This is probably a lesser use case I guess so it might not make as much sense to overoptimize but you could also write a "cheaper" function to just count up the agents and skip some of the other processing.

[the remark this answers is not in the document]
ascibisz: Yeah I don't think it's really worthwhile to keep track of the AgentData object. I just wrote a more lightweight function to just compute the n_agents

blairlyons (approved):

simulariumio/data_objects/simularium_file_data.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self, file_contents: Union[str, bytes]):
+        pass
+
+    def get_frame_at_index(self, frame_number: int) -> Union[FrameData, None]:
+        """
+        Return frame data for frame at index. If there is no frame at the index,
+        return None.
+        """
```
toloudis, line 14: Not sure why I didn't notice this before but are you supposed to put `pass` in each of these functions?  Also it looks like this is supposed/intended tobe an abstract base class 
```
from abc import ABC

class MyABC(ABC):
...
@abstractmethod...
```
Is that sort of the intention here?

simulariumio/data_objects/simularium_file_data.py
```diff
@@ -0,0 +1,42 @@
+from typing import Dict, Union
+from .trajectory_data import TrajectoryData
+from .frame_data import FrameData
+
+
+class SimulariumFileData:
+    def __init__(self, file_contents: Union[str, bytes]):
```
toloudis, line 7: maybe the base class doesn't need to have the file_contents even as a constructor argument?

toloudis (approved):
Overall lgtm, I added a couple of minor comments on the details of the base class but nothing serious.

simulariumio/data_objects/simularium_file_data.py
ascibisz: autoflake doesn't like it when you have `pass` after a docstring, so maybe I'll move the docstrings to the subclasses? I hadn't even noticed that `pass` had been removed, it just disappeared when I ran autoflake locally. Good eye

PR #160, opened 2023-11-02
move things we need in octopus to top level simulariumio package

ascibisz:
Addressing this [comment](https://github.com/simularium/octopus/pull/42#discussion_r1379340629) in my octopus PR so that octopus will look nicer

### DIFF
--- a/simulariumio/__init__.py
+++ b/simulariumio/__init__.py
@@ -22,9 +22,13 @@ from .data_objects import (  # noqa: F401
     ScatterPlotData,
     TrajectoryData,
     UnitData,
+    FrameData,
+    SimulariumFileData,
+    JsonData,
+    BinaryData,
 )
 # DO NOT ISORT DISPLAY_TYPE, CAUSES CIRCULAR DEP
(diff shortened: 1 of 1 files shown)


meganrm (approved):
nice

interim17 (approved):

simulariumio/__init__.py
```diff
@@ -22,9 +22,13 @@
...
     TrajectoryData,
     UnitData,
+    FrameData,
+    SimulariumFileData,
+    JsonData,
+    BinaryData,
 )
 # DO NOT ISORT DISPLAY_TYPE, CAUSES CIRCULAR DEP
-from .constants import DISPLAY_TYPE  # noqa: F401
+from .constants import BINARY_SETTINGS, DISPLAY_TYPE  # noqa: F401
```
toloudis, line 31: this looks like an implementation detail and makes me wonder if BINARY_SETTINGS should live inside of BinaryData.  Just seeing this in isolation so not sure if that would make sense

PR #163, opened 2023-12-01
Update file_format.md

meganrm:
we're on v3 of trajectory file info

### DIFF
--- a/file_format.md
+++ b/file_format.md
@@ -6,7 +6,7 @@
 
 Both JSON and binary contain the following data structured like:
 * **trajectory info**
-  * version - 2.0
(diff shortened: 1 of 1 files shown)


ShrimpCryptid (approved):

interim17 (approved):

toloudis:
What is the difference between v2 and v3?

meganrm:
> What is the difference between v2 and v3? I guess this document is not meant to capture the history of the version differences...



> What is the difference between v2 and v3? I guess this document is not meant to capture the history of the version differences...

It has the optional modelInfo section. The modelInfo was already added in this doc

Issue #109, opened 2022-07-14
Subpoints relative to agent transform


blairlyons:
## Use Case
As a developer, I want spatial values saved in subpoints to be relative to the agent's transform.

## Acceptance Criteria
Subpoints for fibers and sphere groups are relative to the agent's position, rotation, and scale.

## Details
Currently objects with subpoints are given a zero position and rotation so subpoints are (trivially) saved in local space, ideally the main object position would be at the center of the subpoints and the subpoints would be in local space relative to that.

We also need to make sure we're handling rotations properly (e.g. in translate filter), there may need to be some quaternion math for that. Although generally we would set rotations to zero unless the user provides it, and then it's up to them to provide subpoints relative to that correctly.

This needs to be added to documentation.

This will make it easier to transform the agent or focus on it in the viewer code.

PR #169, opened 2024-02-08
subpoints relative to agent position

ascibisz:
Problem
=======
A while ago Dan did some updates to the `simularium-viewer` for subpoints, allowing the agent position / rotation to actually transform the fiber positioning specified in the subpoints (see [here](https://github.com/simularium/simularium-viewer/pull/333)). This PR is applying that to simulariumio's translate filter, so now we'll translate the agent position instead of all the subpoint positions for fibers.

[Link to story or ticket](https://github.com/simularium/simulariumio/issues/109)

Solution
========
- Updated the translate filter and centering logic to translate agent position positions instead of subpoint positions
- Updated the tests accordingly


simulariumio/trajectory_converter.py
```diff
@@ -168,38 +168,9 @@ def translate_positions(data: AgentData, translation: np.ndarray) -> AgentData:
             XYZ translation
         """
         total_steps = data.times.size
```
blairlyons, line 170: call the translate filter here instead of re-implementing it? I forgot we had this redundant method

file_format.md
```diff
@@ -64,7 +64,7 @@ Both JSON and binary contain the following data structured like:
       * subpoints - the number of proceeding values that are extra data belonging to this agent, followed by a list of numerical data
         * for default type (1000), this is not used
           * ex: subpoints = 0
-        * for fiber type (1001), this is the list of positions XYZ of the points along the fiber
+        * for fiber type (1001), this is the list of positions XYZ of the points along the fiber, relative to agent's position specified above
```
blairlyons, line 67: 💯

blairlyons (approved):

blairlyons:
sorry I just remembered we also should save fibers with the center as the position. This will make it so when you click on the fiber the camera centers around the location of the fiber rather than the origin of the scene. 

So for converters that deal with fibers (cellpack, cytosim, medyan, springsalad) or sphere groups (physicell), can you calculate the center of the subpoints in 3D (take mean along each dimension), set that as the agent position, and subtract it from the subpoint positions? lmk if it would help to chat about it

simulariumio/filters/translate_filter.py
```diff
@@ -67,28 +66,5 @@ def apply(self, data: TrajectoryData) -> TrajectoryData:
...
-                        n_items = sp_items.shape[0]
-                        for item_index in range(n_items):
-                            sp_items[item_index][:VALUES_PER_3D_POINT] += translation
-                        n_sp = int(data.agent_data.n_subpoints[time_index][agent_index])
-                        data.agent_data.subpoints[time_index][agent_index][
-                            :n_sp
-                        ] = sp_items.reshape(n_sp)
-                if not translate_subpoints:
-                    # translate agent position for non-fibers
-                    data.agent_data.positions[time_index][agent_index] += translation
```
interim17, line 93: Checking my understanding: Dan's updates to the viewer make this redundant because we add translation and rotation in the shader now

ascibisz: So prior to Dan's changes, the agent position data and rotation data would be ignored for fibers, the viewer would only pay attention to the subpoint position data. With Dan's changes, the position data and rotation data are now taken into account in the viewer, but if the .simularium trajectory has the rotation and position data zeroed out, then we don't even tap into the work Dan did. This change set will allow future trajectories converted with simulariumio to take advantage of these changes Dan made in the viewer!

simulariumio/tests/converters/test_cytosim_converter.py
ShrimpCryptid, line 1: What were the changes in this and the other test files for?

ShrimpCryptid (approved):
Changes are sensible and easy to follow. LGTM!

simulariumio/tests/converters/test_cytosim_converter.py
ascibisz: This was a result of a change Blair recommended in a comment on this PR. For converters that deal with fibers, now rather than set the fiber position just in the subpoints, we also calculate the center of the subpoints, set that as the agent position, and subtract that from the subpoint positions accordingly. This will allow it so that in the viewer, you click on the fiber the camera centers around the location of the fiber appropriately.

So for the tests, the agent XYZ position coordinates should be the average of all it's subpoint XYZ coordinates, and then those values should be subtracted from the subpoints. The changes to the floats reflect that!

ascibisz:
> sorry I just remembered we also should save fibers with the center as the position. This will make it so when you click on the fiber the camera centers around the location of the fiber rather than the origin of the scene.
> 
> So for converters that deal with fibers (cellpack, cytosim, medyan, springsalad) or sphere groups (physicell), can you calculate the center of the subpoints in 3D (take mean along each dimension), set that as the agent position, and subtract it from the subpoint positions? lmk if it would help to chat about it

@blairlyons Ok I have addressed this! The Physicell converter already handled this, since sphere groups subpoints have been relative to agent position from the start, so I didn't need to change anything there :)

simulariumio/trajectory_converter.py
ascibisz: the translate filter expects the full `TrajectoryData` object to be fully formed when it's called, which isn't quite true when we're calling this function from various converters as the `TrajectoryData` object is getting formed. But redundant code is bad, so I moved the redundant functionality to a util function that both places can access to handle that!

blairlyons (approved):

Issue #193, opened 2025-04-18
Rescaling outputs from Cytosim


ferrinm:
## System and Software
* simulariumio Version: 1.12.0
* Python Version: 3.13.2
* Operating System: Ubuntu 23.04

## Description
The `scale_factor` of `MetaData` within `CytosimData` does not behave as expected. When I set `scale_factor=1.0` my data appears correctly oriented in the Simularium viewer, but setting any other value or leaving it to be automatically determined results in different objects being scaled differently from each other and possibly translated relatively to each other as well.

## Expected Behavior
I expected setting a `scale_factor` > 1.0 to make all objects increase in scale by the same amount, so that everything visually looks the same in the viewer but I just don't have to zoom in as far to see it clearly.

## Reproduction
Using the attached Cytosim outputs (point `output_dir` to the directory they are downloaded to in order to reproduce), run the Simularium conversion with `scale_factor` = 1.0 or 2.0. The first looks fine, the second has different objects scaled inconsistently.

[singles.txt](https://github.com/user-attachments/files/19811585/singles.txt)
[couples.txt](https://github.com/user-attachments/files/19811586/couples.txt)
[solids.txt](https://github.com/user-attachments/files/19811587/solids.txt)
[fiber_points.txt](https://github.com/user-attachments/files/19811584/fiber_points.txt)

```
import os
import numpy as np
from simulariumio.cytosim import CytosimConverter, CytosimData, CytosimObjectInfo
from simulariumio import MetaData, DisplayData, DISPLAY_TYPE, ModelMetaData, InputFileData

output_dir = #PATH TO CYTOSIM OUTPUTS

for scale_factor in [1.0,2.0]:
    example_data = CytosimData(
        meta_data=MetaData(
            box_size=np.array([1.0,1.0,0.8]),
            scale_factor=scale_factor,
        ),
        object_info={
            "fibers" : CytosimObjectInfo(
                cytosim_file=InputFileData(
                    file_path=os.path.join(output_dir,"fiber_points.txt"),
                ),
                display_data={
                    0 : DisplayData(
                        name="actin",
                        radius=0.005,
                        display_type=DISPLAY_TYPE.FIBER
                    )
                },
                position_indices=[1, 2, 3]
            ),
            "solids" : CytosimObjectInfo(
                cytosim_file=InputFileData(
                    file_path=os.path.join(output_dir,"solids.txt"),
                ),
                display_data={
                    0 : DisplayData(
                        name="pit",
[...]

PR #194, opened 2025-04-24
Cytosim Scaling Bug

ascibisz:
Time estimate or Size
=======
small

Problem
=======
A user reported that scaling wasn't working as expected for the Cytosim converter. 

Looking at the code, we were erroneously scaling all agent data that had been read thus far with each file read. So for example, for the user that reported the bug, there were 4 input files, so the first file that was read would get scaled by the scale factor 4 times, the second would get scaled 3 times, the third would get scaled twice, and the final file would get scaled the expected one time. This resulted in things looking very weird and unexpected!

[Link to story or ticket](https://github.com/simularium/simulariumio/issues/193)

Solution
========
Let's just apply the scale factor once per conversion, after we have all the data read in. This is a very small change, just  moving the scale factor part to be after the for loop where we're going through the input files.

Also let's update the test to make sure this doesn't happen again. Our original scale factor test only read in one file, so thus the problem with multiple files was missed.

### DIFF
--- a/simulariumio/cytosim/cytosim_converter.py
+++ b/simulariumio/cytosim/cytosim_converter.py
@@ -191,7 +191,6 @@ class CytosimConverter(TrajectoryConverter):
         used_unique_IDs: List[int],
         overall_line: int,
         total_lines: int,
-        scale_factor: float = None,
     ) -> Tuple[Dict[str, Any], List[int], int]:
         """
         Parse a Cytosim output file containing objects
@@ -274,12 +273,9 @@ class CytosimConverter(TrajectoryConverter):
                 )
                 result.n_agents[time_index] += 1
             self.check_report_progress(overall_line / total_lines)
-        result, scale_factor = TrajectoryConverter.scale_agent_data(
-            result, scale_factor
-        )
         result = TrajectoryConverter.center_fiber_positions(result)
         result.n_timesteps = time_index + 1
-        return (result, used_unique_IDs, overall_line, scale_factor)
+        return (result, used_unique_IDs, overall_line)
 
     def _read(self, input_data: CytosimData) -> TrajectoryData:
         """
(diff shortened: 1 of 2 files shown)


toloudis (approved):
LGTM!  Thank you for the quick fix.  Too bad we don't have "simpler" test data so it's easier to see how the tests work.

simulariumio/tests/converters/test_cytosim_converter.py
```diff
@@ -118,7 +118,7 @@ def test_box_size_provided(box_size, expected_box_size):
 
 # value of automatically generated scale factor, so that position
 # data fits within VIEWER_DIMENSION_RANGE
-range = 0.400052 - -0.001
+range = 0.00808633 - -0.39575567
```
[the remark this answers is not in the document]
ascibisz: I'm not sure how I came up with the other numbers previously, but the new ones are the actual min / max numbers that define the range. I just noticed this as I was testing, since I was also in the weeds on testing the autoscaling

rugeli (approved):

simulariumio/cytosim/cytosim_converter.py
```diff
@@ -191,7 +191,6 @@ def _parse_objects(
         used_unique_IDs: List[int],
         overall_line: int,
         total_lines: int,
-        scale_factor: float = None,
```
rugeli, line 194: Just curious: what's deciding the scale_factor should be a float or not

ascibisz: So scale_factor is an optional input parameter that the user may or may not provide when setting up their `CytosimData` object.  Then, in `TrajectoryConverter.scale_agent_data()`, which I didn't change in this PR, if scale_factor is None, we generate a scale factor to get the data to look decent in the viewer (so like not be so tiny you can't see anything or so huge that it's out of frame).

PR #195, opened 2025-05-02
Mem3DG converter

ascibisz:
Time estimate or Size
=======
Medium-ish

Problem
=======
Creating a new converter for Mem3DG output data
[Link to story or ticket](https://github.com/simularium/simulariumio/issues/166)

Solution
========
This converter ended up different than the normal simulariumio converters. Mem3DG creates a single .nc file, which in all the examples I successfully ran looks like a mesh object that changes across the timesteps. 

The best / only way I came up with to bring this into simularium was to take the mesh data at each time stamp and convert that into a .obj mesh file, which simularium knows how to render. Then, generate a simularium file that references each of these generated mesh files.

This does work, but we have these pieces that are different than the "normal" simulariumio / simularium pattern
1. Rather than generating a single .simularium file via simulariumio, this generates one .simularium files and as many .obj files as there are timesteps
2. To get the trajectory to load in simularium with the files locally, you have to import the .simularium file and all of the .obj files
3. Each time step is a different agent, which visually isn't too bad since they'll just show up as subtypes of the same original type, but it's still different than usual

Also one outstanding question I have not yet answered: is there a situation in which a single .nc file would represent multiple distinct meshes? This converter code currently assumes one .nc file -> 1 mesh agent per timestep -> 1 .obj file generated per timestep.

## Type of change

Change summary:
---------------
* Added the Mem3dg converter, and all the things that tend to come with a new converter (`Mem3dgData`, unit tests, and jupyter notebook example)

Steps to Verify:
----------------
The best way to test this would be by checking out this branch and trying the Jupyter Notebook tutorial file! If it anything doesn't work or if it's confusing, that is all helpful feedback :)


simulariumio/mem3dg/mem3dg_converter.py
```diff
@@ -0,0 +1,104 @@
...
+
+            agent_data.radii[frame][0] = input_data.meta_data.scale_factor
+            agent_data.unique_ids[frame][0] = frame
+
+            name = str(frame)
+            agent_data.types[frame].append(name)
+            object_display_data = DisplayData(
+                name=f"{base_agent_name}#frame{frame}",
+                display_type=DISPLAY_TYPE.OBJ,
+                url=f"{frame}.obj",
```
rugeli, line 81: nice job fixing the path issue! The different between `url` and `output_file_path` isn't immediately clear, perhaps rename `url` to clarify?

simulariumio/mem3dg/mem3dg_converter.py
```diff
@@ -0,0 +1,104 @@
...
+                file.write(f"f {t[0]} {t[1]} {t[2]}\n")
+
+    def _read_traj_data(self, input_data: Mem3dgData) -> AgentData:
+        try:
+            data = Dataset(input_data.input_file_path, "r")
+            n_frames = np.size(data.groups["Trajectory"].variables["time"])
+        except Exception as e:
+            raise InputDataError(f"Error reading input Mem3DG data: {e}")
+
+        # for now, we are representing converted Mem3DG trajectoies as one
```
interim17, line 59:
```suggestion
        # for now, we are representing converted Mem3DG trajectories as one
```

interim17 (approved):
LGTM, haven' run it.

The multiple .obj to one simualarium file pattern is interesting to me. It reminds me a bit of .zarr and seems like a way we could approach have different levels of resolution in a simularium file going forward if and when we take a look at the file structure or aim to work on that feature.

PR #201, opened 2025-08-14
Update print statements to use logging in `json_writer.py`

mogres:
Time estimate or Size
=======
xsmall

Problem
=======
cellPACK uses simulariumio to convert outputs. The output stream gets a lot of noisy outputs during the conversion process. Changing the print statements to logging will allow users to configure the output level as needed.

Solution
========
Changes print statements to use python logging in `json_writer.py`

## Type of change

Change summary:
---------------
* print-> log.info


toloudis (approved):
Good idea, LGTM.

PR #206, opened 2026-03-31
Add USD to Simularium converter

dillonl:
Time estimate or Size
Small: ~1 hour to review. New self-contained module with no changes to existing code paths.

Problem
There is no way to convert USD (Universal Scene Description) files into the .simularium format. USD is a widely used interchange format exported by tools like Blender, Maya, and Houdini, and is a natural input for biological simulation visualization workflows.

Solution
Added a UsdConverter and UsdData class following the same patterns as existing converters (e.g. CellPackConverter). The converter:

Opens a USD stage and discovers all mesh geometry prims
Deduplicates meshes by geometry hash and normalizes them relative to their local origin (preserving pivot points set in DCC tools)
Extracts per-frame local transforms using Gf.Transform, decomposing rotation into intrinsic XYZ Euler angles compatible with the THREE.js viewer
Applies np.unwrap per-axis to remove Euler discontinuities
Supports optional animation trimming (trim_to_animation=True by default) to crop to the actual animated frame range
Supports display data overrides by name
Writes mesh geometry as .obj files alongside the .simularium output

Type of change
New feature (non-breaking change which adds functionality)
This change requires updated or new tests

Change summary:
Add USD converter for importing Universal Scene Description files
Fix USD rotation convention to match THREE.js Euler('XYZ') intrinsic order
Clean up USD converter: remove stale comments, unused imports, and improve error handling

Steps to Verify:
Install with USD support: pip install ".[usd]"
Run the test suite: pytest simulariumio/tests/converters/test_usd_converter.py
Convert a USD file manually:

from simulariumio.usd import UsdConverter, UsdData
converter = UsdConverter(UsdData(usd_file_path="my_file.usd"))
converter.save("output")
Open output.simularium in the Simularium viewer at simularium.allencell.org
Verify agents appear with correct positions, rotations, and mesh shapes across frames

Keyfiles:
simulariumio/usd/usd_converter.py -> main converter logic
simulariumio/usd/usd_data.py -> input data container
simulariumio/tests/converters/test_usd_converter.py -> test suite


simulariumio/usd/usd_converter.py
```diff
@@ -0,0 +1,339 @@
...
+        """
+        print("Reading USD Data -------------")
+        try:
+            stage = Usd.Stage.Open(input_data.usd_file_path)
+        except Exception:
+            stage = None
+        if stage is None:
+            raise FileNotFoundError(
+                f"Could not open USD file: {input_data.usd_file_path}"
+            )
```
interim17, line 153: Possible to surface an error msg here instead of raising generic FileNotFound?

something like:
```
        except Exception as e:
            raise InputDataError(f"Error reading input USD data: {e}")
```

simulariumio/usd/usd_converter.py
```diff
@@ -0,0 +1,339 @@
...
+        self._mesh_to_obj = {}
+        super().__init__(input_data, progress_callback, callback_interval)
+        self._data = self._read(input_data)
+
+    @staticmethod
+    def _geometry_hash(points, face_vertex_indices) -> str:
+        """Compute a hash for mesh geometry to identify duplicates."""
+        pts_bytes = np.array(points).round(6).tobytes()
+        idx_bytes = np.array(face_vertex_indices, dtype=np.int32).tobytes()
+        return hashlib.sha256(pts_bytes + idx_bytes).hexdigest()
```
interim17, line 70: Mesh dedup is nice!

interim17 (commented):
Overall this is looking great! Thank you for your efforts.

A couple questions.

1. At first pass running with this, the USDascii test file converted into a simularium file with 340 time steps, and my conversion of USDbinary test data resulted in a single frame (looked like the last frame of the USDascii trajectory). Would you expect them to be the same? Possible user error on my end?

2. For CI would be good to add `usd` to the array on line 26 of Justfile

dillonl:
> Overall this is looking great! Thank you for your efforts.
> 
> A couple questions.
> 
> 1. At first pass running with this, the USDascii test file converted into a simularium file with 340 time steps, and my conversion of USDbinary test data resulted in a single frame (looked like the last frame of the USDascii trajectory). Would you expect them to be the same? Possible user error on my end?
> 2. For CI would be good to add `usd` to the array on line 26 of Justfile

1. Good eye, this is expected behavior, not a bug. The two test fixtures aren't equivalent: actin_USDascii.usd has 50 keyframes over a declared 1–400 frame range, while actin_USDbinary.usd was exported as a static snapshot (start/end time code both 0.0, zero time samples). So the binary file legitimately collapses to a single frame. If you'd like, I can regenerate the binary fixture from the same animated source so they round-trip to the same trajectory, just let me know.

2. As for the Justfile, I added usd to the install extras in the Justfile

simulariumio/usd/usd_converter.py
dillonl: Good call, I switched to InputDataError with a descriptive message, matching the pattern you suggested. Both the explicit None return and any exception from Usd.Stage.Open now surface as InputDataError("Error reading input USD data: ...").

simulariumio/usd/usd_converter.py
dillonl: Thanks!

simulariumio/tests/converters/test_usd_converter.py
ascibisz, line 1: Thanks for writing all these tests! Would it be possible to add a test where `trim_to_animation=True`? That's one of the paths through your code that I'm not sure how to test locally, since I've never actually generated a usd file myself 😅

dillonl: Good thinking, I added two tests under TestUsdTrimToAnimation. The ascii fixture declares an end time code of 400 but the last keyed frame across all agents/ops is 340, so the tests verify that trim_to_animation=True produces 340 frames (vs. 400 untrimmed) and that the kept frames contain identical positions/rotations to the untrimmed run (i.e. we're only dropping the held tail, not altering animation data).

ascibisz (approved):
Everything looks good to me! I was able to convert the test files to simularium file format using your converter without issue and the output files looked and behaved as I expected, nicely done

interim17 (approved):
Nice work! Thanks for the contribution.

simulariumio/tests/converters/test_usd_converter.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def test_obj_has_correct_geometry(self, ascii_converter, tmp_path):
+        output = str(tmp_path / "test_output")
+        ascii_converter.save(output, binary=True)
+
+        obj_path = tmp_path / "mesh_0.obj"
+        with open(obj_path) as f:
+            lines = f.readlines()
+
+        v_lines = [l for l in lines if l.startswith("v ")]
```
toloudis:
```suggestion
        v_lines = [lv for lv in lines if lv.startswith("v ")]
```

simulariumio/tests/converters/test_usd_converter.py
```diff
@@ -0,0 +1,267 @@
...
+        f_lines = [l for l in lines if l.startswith("f ")]
```
toloudis:
```suggestion
        f_lines = [lf for lf in lines if lf.startswith("f ")]
```

PR #208, opened 2026-04-01
Fix writing to read-only array bug in nerdss_converter.py

PhysFoley:
This pull request fixes a small bug in the NERDSS converter that was causing it to fail due to attempting to write to a read-only numpy array.

Time estimate or Size
=======
xsmall

Problem
=======
NERDSS trajectories were failing to convert to simularium format due to an issue within the NerdssConverter class where it tried to write to elements of a read-only numpy array.

Solution
========
Create a copy of the numpy array with the same shape, edit this array, then overwrite the original.

## Type of change
- Bug fix

Change summary:
---------------
* Fixed error "assignment destination is read-only" when calling NerdssConverter._read_pdb_files()

Steps to Verify:
----------------
1. Attempt to convert a NERDSS simulation using [ionerdss.convert_simularium()](https://github.com/JohnsonBiophysicsLab/ionerdss/blob/main/tutorials/ionerdss_tutorial_simularium.ipynb)

Keyfiles (delete if not relevant):
-----------------------
1. simulariumio/nerdss/nerdss_converter.py

### DIFF
--- a/simulariumio/nerdss/nerdss_converter.py
+++ b/simulariumio/nerdss/nerdss_converter.py
@@ -1,6 +1,8 @@
 from MDAnalysis import Universe
 import os
 
+import numpy as np
+
 from ..trajectory_converter import TrajectoryConverter
 from ..data_objects import (
     AgentData,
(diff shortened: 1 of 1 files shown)


toloudis (approved):
Looks good to me.

interim17 (approved):

toloudis:
this seems to have broken linting: https://github.com/simularium/simulariumio/actions/runs/24211560942

PR #211, opened 2026-05-05
Fix linting errors from prior PR #208

PhysFoley:
This is just a PR to fix the [linter errors](https://github.com/simularium/simulariumio/pull/208#issuecomment-4373179423) from my already-merged previous PR #208 

Apologies for not catching these before submitting the PR. I am unable to get the linter running locally. Several tries in fresh environments always produce some kind of crash when running "just lint" that seems to originate from a wheel that fails to build.

I used the output of the linter [here](https://github.com/simularium/simulariumio/actions/runs/24211560942/job/70681398416#step:6:31) to fix the offending lines.

### DIFF
--- a/simulariumio/nerdss/nerdss_converter.py
+++ b/simulariumio/nerdss/nerdss_converter.py
@@ -136,7 +136,7 @@ class NerdssConverter(TrajectoryConverter):
 
         # need our own write-able n_subpoints array
         n_subpoints = np.zeros_like(agent_data.n_subpoints)
-        
+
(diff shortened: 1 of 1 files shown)


toloudis:
Thanks so much for addressing it. We should have cleaned it up before merging it so thanks for the additional PR.

toloudis (approved):
👍

meganrm (approved):

rugeli (approved):

